Issue #66, opened 2023-06-15
`AddFile()` allows also adding directory. Make that clear or add another function


smuu:

Ehsanino82:
https://github.com/celestiaorg/knuu/pull/86
i made this pr for this issue.

PR #77, opened 2023-06-16
feat: add folder function

smuu:
For clarity added a `AddFolder()` function.

Ref.: https://github.com/celestiaorg/knuu/issues/55

## Checklist

- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

PR #86, opened 2023-06-17
change AddFile and add Add function

Ehsanino82:
## Checklist

### DIFF
--- a/pkg/knuu/instance.go
+++ b/pkg/knuu/instance.go
@@ -242,9 +242,9 @@ func (i *Instance) ExecuteCommand(command ...string) (string, error) {
 	return "", nil
 }
 
-// AddFile adds a file to the instance
+// Add adds to the instance
 // This function can only be called in the state 'Preparing'
-func (i *Instance) AddFile(src string, dest string, chown string) error {
+func (i *Instance) Add(src string, dest string, chown string) error {
 	if !i.IsInState(Preparing) {
 		return fmt.Errorf("adding file is only allowed in state 'Preparing'. Current state is '%s'", i.state.String())
 	}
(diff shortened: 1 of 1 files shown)


Ehsanino82:
https://github.com/celestiaorg/knuu/issues/66
handle this issue

pkg/knuu/instance.go
```diff
@@ -258,11 +258,18 @@ func (i *Instance) AddFile(src string, dest string, chown string) error {
...
+	return i.AddFile(src, dest, chown)
+}
+
 // AddFolder adds a folder to the instance
 // This function can only be called in the state 'Preparing'
 func (i *Instance) AddFolder(src string, dest string, chown string) error {
-	// As dockers `ADD` works for both files and folders, we can just call AddFile here
-	return i.AddFile(src, dest, chown)
+	// As dockers `ADD` works for both files and folders, we can just call Add here
+	return i.Add(src, dest, chown)
```
jrmanes, line 272: if they do the same because of docker, why keep both functions?, we could have only `Add` and do it there, wdyt?

jrmanes:
thanks for the PR, very appreciated! 😃

pkg/knuu/instance.go
```diff
@@ -258,11 +258,18 @@ func (i *Instance) AddFile(src string, dest string, chown string) error {
 	return nil
 }
 
+// AddFile adds a file to the instance
+// This function can only be called in the state 'Preparing'
+func (i *Instance) AddFile(src string, dest string, chown string) error {
+	// As dockers `ADD` works for both files and folders, we can just call Add here
+	return i.AddFile(src, dest, chown)
```
smuu, line 265: Recursive call, should be:

```suggestion
	return i.Add(src, dest, chown)
```

pkg/knuu/instance.go
smuu: I like to keep the separate functions as the logic of adding folders and files might change. I already thought of checking whether the files the developer wants to add exist. This might differ between folders and files, idk.

pkg/knuu/instance.go
```diff
@@ -242,9 +242,9 @@ func (i *Instance) ExecuteCommand(command ...string) (string, error) {
 	return "", nil
 }
 
-// AddFile adds a file to the instance
+// Add adds to the instance
 // This function can only be called in the state 'Preparing'
-func (i *Instance) AddFile(src string, dest string, chown string) error {
+func (i *Instance) Add(src string, dest string, chown string) error {
```
smuu, line 247: Currently, I'm not sure if I like the function name `Add()`. Because from an Instance view, what does `add` mean? An Instance abstracts Docker and Kubernetes, and it's unclear that `add` means adding a file/directory.

smuu: Also, I created https://github.com/celestiaorg/knuu/pull/77 to resolve https://github.com/celestiaorg/knuu/issues/66, but I forgot to close the issue. I will keep this PR open for now and decide together with @jrmanes how we want to move forward regarding this topic.

smuu (commented):
Thanks for the contribution :rocket:

smuu:
Superseded by https://github.com/celestiaorg/knuu/pull/77

Issue #45, opened 2023-05-23
Deduplicate code - cloning


smuu:
https://github.com/celestiaorg/knuu/blob/175e6abf10f47371bd59f819a93037254f0b9772/pkg/knuu/instance_helper.go#L172
https://github.com/celestiaorg/knuu/blob/175e6abf10f47371bd59f819a93037254f0b9772/pkg/knuu/instance.go#L384

PR #87, opened 2023-06-17
remove duplication

Ehsanino82:
Resolves https://github.com/celestiaorg/knuu/issues/45

## Checklist

### DIFF
--- a/pkg/knuu/instance.go
+++ b/pkg/knuu/instance.go
@@ -598,6 +598,5 @@ func (i *Instance) Clone() (*Instance, error) {
 	// Create a new instance with the same attributes as the original instance
 	ins := i.cloneWithSuffix("")
 	ins.k8sName = newK8sName
-	ins.instanceType = i.instanceType
 	return ins, nil
 }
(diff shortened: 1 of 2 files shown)


Ehsanino82:
https://github.com/celestiaorg/knuu/issues/45
handle this issue

pkg/knuu/instance.go
```diff
@@ -596,24 +596,8 @@ func (i *Instance) Clone() (*Instance, error) {
...
-		args:              i.args,
-		env:               i.env,
-		volumes:           i.volumes,
-		memoryRequest:     i.memoryRequest,
-		memoryLimit:       i.memoryLimit,
-		cpuRequest:        i.cpuRequest,
-	}, nil
+	ins := i.cloneWithSuffix("")
+	ins.k8sName = newK8sName
+	ins.instanceType = i.instanceType
```
smuu, line 601: Instead of setting this here, this should be part of the `cloneWithSuffix` function. I just saw that I missed adding that to the struct creation there, which is a bug. Good catch! Could you remove the line here and add it to the struct creation in `cloneWithSuffix`?

smuu (changes requested):
Looks good! Thanks for your contribution :rocket:

Only one small thing, and then we can merge your PR :)

pkg/knuu/instance.go
Ehsanino82: yes of course. i did it.

Ehsanino82:
> Looks good! Thanks for your contribution rocket
> 
> Only one small thing, and then we can merge your PR :)

I apply the change you want,

smuu (approved):

Issue #84, opened 2023-06-16
Add a timeout variable


smuu:
Ref.: https://github.com/celestiaorg/knuu/issues/83

PR #110, opened 2023-06-23
Feature/global timeout

smuu:
Introduces a new env variable `KNUU_TIMEOUT` with the default of `1h`, which will delete all resources that are still running after that time for that specific test run.

Resolves https://github.com/celestiaorg/knuu/issues/84, resolves https://github.com/celestiaorg/knuu/issues/78, resolves https://github.com/celestiaorg/knuu/issues/18

## Checklist

### DIFF
--- a/pkg/k8s/k8s_rolebinding.go
+++ b/pkg/k8s/k8s_rolebinding.go
@@ -1,0 +1,60 @@
+package k8s
+
+import (
+	"context"
+	"fmt"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
+)
+
+// CreateRoleBinding creates a roleBinding
+func CreateRoleBinding(name, namespace string, labels map[string]string, role, serviceAccount string) error {
+
+	roleBinding := &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
+		},
+		RoleRef: rbacv1.RoleRef{
+			Kind: "Role",
+			Name: role,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      serviceAccount,
+				Namespace: namespace,
+			},
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+
+	if !IsInitialized() {
+		return fmt.Errorf("knuu is not initialized")
+	}
+	if _, err := Clientset().RbacV1().RoleBindings(namespace).Create(ctx, roleBinding, metav1.CreateOptions{}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// DeleteRoleBinding deletes a roleBinding
+func DeleteRoleBinding(name, namespace string) error {
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+
+	if !IsInitialized() {
+		return fmt.Errorf("knuu is not initialized")
+	}
+	if err := Clientset().RbacV1().RoleBindings(namespace).Delete(ctx, name, metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/k8s/k8s_serviceaccount.go
+++ b/pkg/k8s/k8s_serviceaccount.go
@@ -1,0 +1,49 @@
+package k8s
+
+import (
+	"context"
+	"fmt"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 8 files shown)


pkg/k8s/k8s_rolebinding.go
```diff
@@ -0,0 +1,60 @@
...
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      serviceAccount,
+				Namespace: namespace,
+			},
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
Bidon15, line 33: you can make 20*time.Second a const and use it twice

pkg/k8s/k8s_serviceaccount.go
```diff
@@ -0,0 +1,49 @@
...
+
+	serviceAccount := &v1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
Bidon15, line 22: as you are defining a timeout - you can declare the var a pkg const 
here is the same

pkg/knuu/knuu.go
```diff
@@ -54,10 +56,72 @@ func InitializeWithIdentifier(uniqueIdentifier string) error {
...
 }
+
+// handleTimeout creates a timeout handler that will delete all resources with the identifier after the timeout
+func handleTimeout() error {
+
+	instance, err := NewInstance("timeout-handler")
+	if err != nil {
+		return fmt.Errorf("cannot create instance: %s", err)
+	}
+	// FIXME: use supported kubernetes version images (use of latest could break)
```
Bidon15, line 91: make a todo with issue in the repo and link it here
We will remember to do this nit later on

pkg/knuu/knuu.go
```diff
@@ -54,10 +56,72 @@ func InitializeWithIdentifier(uniqueIdentifier string) error {
...
+		return fmt.Errorf("cannot set image: %s", err)
+	}
+	if err := instance.Commit(); err != nil {
+		return fmt.Errorf("cannot commit instance: %s", err)
+	}
+	timeoutSeconds := int64(timeout.Seconds())
+
+	// command to wait for timeout and delete all resources with the identifier
+	var command = []string{"sh", "-c"}
+	cmd := fmt.Sprintf("sleep %d && kubectl delete all,pvc,netpol,roles,serviceaccounts,rolebindings -l test-run-id=%s -n %s --wait=false", timeoutSeconds, identifier, k8s.Namespace())
```
Bidon15, line 102: [non-blocking]: Isn't it better to use the go client api for this? Just curious why you opted for this

Bidon15 (changes requested):
Some small improvements. Otherwise LGTM

pkg/k8s/k8s_rolebinding.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"fmt"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
+)
+
+// CreateRoleBinding creates a roleBinding
+func CreateRoleBinding(name, namespace string, labels map[string]string, role, serviceAccount string) error {
```
jrmanes, line 12: thinking loudly, if we are going to use this function to create rolebinding role<->sa, we could add it to the comments to generate the docs correctly, to prevent any misunderstanding (rolebinding can bind role<->user for example), wdyt

pkg/knuu/knuu.go
smuu: This command runs in a container in the cluster. Not on the machine where the test is executed. To make it as minimal and easy as possible, I just run a simple image like a Pod and run a shell. The alternative would be to create a separate Go application. I was considering it but decided to go for the simple solution.

pkg/k8s/k8s_rolebinding.go
smuu: https://github.com/celestiaorg/knuu/issues/115

pkg/k8s/k8s_serviceaccount.go
smuu: https://github.com/celestiaorg/knuu/issues/115

pkg/knuu/knuu.go
smuu: linked the issue, thanks for that :+1:

pkg/knuu/knuu.go
Bidon15: then this approach sounds sensible to me. Woud you mind putting this as a comment in the cmd var? 🙏

Bidon15 (dismissed):
LGTM

jrmanes (approved):
🚀

Bidon15 (approved):

PR #112, opened 2023-06-23
feat: delete with no grace period when callind destroy to speed up cl…

smuu:
…ean up

Resolves https://github.com/celestiaorg/knuu/issues/101

## Checklist


pkg/knuu/instance_helper.go
```diff
@@ -166,7 +166,8 @@ func (i *Instance) deployPod() error {
 // destroyPod destroys the pod for the instance
 // Skips if the pod is already destroyed
 func (i *Instance) destroyPod() error {
-	err := k8s.DeleteStatefulSet(k8s.Namespace(), i.k8sName)
+	grace := int64(0)
```
Bidon15, line 169: we need a comment for ourselves in the func that now it's a force delete with no grace period

Issue #103, opened 2023-06-22
Add delete function to Preloader


smuu:
call `EmptyImages()`

salehshoji:
is this okay?


```
func NewPreloader() (*Preloader, error) {
	k8sName, err := generateK8sName("knuu-preloader")
	if err != nil {
		return nil, fmt.Errorf("error generating k8s name for preloader: %w", err)
	}
	preloader := &Preloader{
		k8sName: k8sName,
		images:  []string{},
	}
	err = preloader.EmptyImages()
	if err != nil {
		return nil, err
	}
	return preloader, nil
}
```

salehshoji:
#114

smuu:
Thanks for the question :rocket: 

I meant adding `func (p *Preloader) Delete() error`, which in the implementation can call `EmptyImages()`.

PR #114, opened 2023-06-26
call EmptyImage

salehshoji:
#103


smuu:
Thanks for the contribution @salehshoji :rocket: 

I added a comment to the issue. See here: https://github.com/celestiaorg/knuu/issues/103#issuecomment-1623210571

PR #127, opened 2023-07-03
feat: add probes

smuu:
Tested on branch https://github.com/celestiaorg/knuu-example/tree/feature/add-probes

Resolves https://github.com/celestiaorg/knuu/issues/118

## Checklist


pkg/k8s/k8s_pod_probes.go
```diff
@@ -0,0 +1,66 @@
...
+	//Exec                          *ExecAction
+	HTTPGet             *HTTPGetAction   // Only one action is allowed, deployment will fail otherwise
+	TCPSocket           *TCPSocketAction // Only one action is allowed, deployment will fail otherwise
+	InitialDelaySeconds int32            // Optional
+	//TerminationGracePeriodSeconds int64
+	//PeriodSeconds                 int32
+	//TimeoutSeconds                int32
+	//FailureThreshold              int32
+	//SuccessThreshold              int32
+	//GRPC                          *GRPCAction
```
Bidon15, line 20: Can you elaborate on why we need to comment this out?

pkg/k8s/k8s_pod_probes.go
```diff
@@ -0,0 +1,66 @@
...
+//type ExecAction struct {
+//	Command []string
+//}
+
+type HTTPGetAction struct {
+	Port int32
+	//Host        string
+	//HttpHeaders []HTTPHeader
+	Path string
+	//Scheme      string
```
Bidon15, line 42: same here

pkg/k8s/k8s_pod_probes.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (p *Probe) String() string {
+	if p.HTTPGet != nil {
+		return fmt.Sprintf("HTTPGet: %s, InitialDelaySeconds: %d", p.HTTPGet.String(), p.InitialDelaySeconds)
+	}
+	if p.TCPSocket != nil {
+		return fmt.Sprintf("TCPSocket: %s, InitialDelaySeconds: %d", p.TCPSocket.String(), p.InitialDelaySeconds)
+	}
+	return ""
```
Bidon15, line 30: Can we give a hint why empty string is good here?

pkg/knuu/instance.go
```diff
@@ -538,6 +550,42 @@ func (i *Instance) SetServiceAccount(serviceAccount string) error {
 	return nil
 }
 
+// SetLivenessProbe sets the liveness probe of the instance
+// A live probe is a probe that is used to determine if the instance is still alive, and should be restarted if not
+// This function can only be called in the states 'Preparing' and 'Committed'
+func (i *Instance) SetLivenessProbe(livenessProbe *k8s.Probe) error {
+	if !i.IsInState(Preparing, Committed) {
+		return fmt.Errorf("setting liveness probe is only allowed in state 'Preparing' or 'Committed'. Current state is '%s'", i.state.String())
+	}
```
Bidon15, line 559: [non-blocking]: this can be a good first issue to create a private func that does the checking for errors based on inputs of errors

Bidon15 (changes requested):
Some questions around commented code-snippets. Otherwise, it looks good enough 👍

pkg/k8s/k8s_pod_probes.go
jrmanes: compiler I guess

pkg/k8s/k8s_pod_probes.go
smuu: https://github.com/celestiaorg/knuu/pull/127/files#diff-97b6d1c19f51934d9900ee900ffd698b15a1750ea95088d1f30ff73d9a98fb5dR5-R7

pkg/k8s/k8s_pod_probes.go
smuu: https://github.com/celestiaorg/knuu/pull/127/files#diff-97b6d1c19f51934d9900ee900ffd698b15a1750ea95088d1f30ff73d9a98fb5dR5-R7

pkg/k8s/k8s_pod_probes.go
smuu: What would you propose to return here?

pkg/knuu/instance.go
smuu: Do you mean something like this: https://github.com/celestiaorg/knuu/issues/15?

Bidon15: Not really. You can just encapsulate the checker if into a single private func that will be called by 3 Public funcs providing it (private) with strings each

pkg/k8s/k8s_pod_probes.go
Bidon15: I would return something meaningful like 'Probing TCP and HTTP sockets has recorded zero delays'. 

I don't see "" being useful for a human being. And for code-base further logic, we should refer to int InitialDelaySeconds rather then String()

pkg/k8s/k8s_pod_probes.go
Bidon15: This is not a good explanation and the way to go tbh. I'd propose you start with Interfacing to achieve duck typing and leave room for future improvements with backwards compat. 

Rather than off-setting the choice to the user to uncomment and do the implementation

Bidon15 (approved):
LGTM

PR #129, opened 2023-07-03
feat: add rbac

smuu:
```[tasklist]
### TODO
- [x] Test with a separate Instance, not only with the timeout handler
```

Tested on branch https://github.com/celestiaorg/knuu-example/tree/feature/add-rbac

Resolves https://github.com/celestiaorg/knuu/issues/119

## Checklist

### DIFF
--- a/pkg/knuu/instance.go
+++ b/pkg/knuu/instance.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	appv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"os"
 	"path/filepath"
 	"strings"
@@ -33,7 +34,7 @@ type Instance struct {
 	memoryRequest         string
 	memoryLimit           string
 	cpuRequest            string
-	serviceAccountName    string
+	policyRules           []rbacv1.PolicyRule
 	livenessProbe         *v1.Probe
 	readinessProbe        *v1.Probe
 	startupProbe          *v1.Probe
@@ -49,24 +50,24 @@ func NewInstance(name string) (*Instance, error) {
 	}
 	// Create the instance
 	return &Instance{
-		name:               name,
-		k8sName:            k8sName,
-		imageName:          "",
-		state:              None,
-		instanceType:       BasicInstance,
-		portsTCP:           make([]int, 0),
-		portsUDP:           make([]int, 0),
-		command:            make([]string, 0),
-		args:               make([]string, 0),
-		env:                make(map[string]string),
-		volumes:            make([]*k8s.Volume, 0),
-		memoryRequest:      "",
-		memoryLimit:        "",
-		cpuRequest:         "",
-		serviceAccountName: "default",
-		livenessProbe:      nil,
-		readinessProbe:     nil,
-		startupProbe:       nil,
+		name:           name,
+		k8sName:        k8sName,
+		imageName:      "",
+		state:          None,
+		instanceType:   BasicInstance,
+		portsTCP:       make([]int, 0),
+		portsUDP:       make([]int, 0),
+		command:        make([]string, 0),
+		args:           make([]string, 0),
+		env:            make(map[string]string),
+		volumes:        make([]*k8s.Volume, 0),
+		memoryRequest:  "",
+		memoryLimit:    "",
+		cpuRequest:     "",
+		policyRules:    make([]rbacv1.PolicyRule, 0),
+		livenessProbe:  nil,
+		readinessProbe: nil,
+		startupProbe:   nil,
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/knuu/instance_helper.go
```diff
@@ -124,6 +124,23 @@ func (i *Instance) deployPod() error {
...
+
+	// create a role and role binding for the pod if there are policy rules
+	if len(i.policyRules) > 0 {
+		if err := k8s.CreateRole(k8s.Namespace(), i.k8sName, labels, i.policyRules); err != nil {
+			return fmt.Errorf("failed to create role: %v", err)
+		}
+		if err := k8s.CreateRoleBinding(k8s.Namespace(), i.k8sName, labels, serviceAccountName, i.k8sName); err != nil {
+			return fmt.Errorf("failed to create role binding: %v", err)
+		}
+	}
```
Bidon15, line 142: I am confused. Why do we want to use k8sName as a serviceAccountName and remove instance's serviceAccountName variable?

pkg/knuu/instance.go
```diff
@@ -45,21 +46,21 @@ func NewInstance(name string) (*Instance, error) {
...
+		portsTCP:      make([]int, 0),
+		portsUDP:      make([]int, 0),
+		command:       make([]string, 0),
+		args:          make([]string, 0),
+		env:           make(map[string]string),
+		volumes:       make([]*k8s.Volume, 0),
+		memoryRequest: "",
+		memoryLimit:   "",
+		cpuRequest:    "",
+		policyRules:   make([]rbacv1.PolicyRule, 0),
```
jrmanes, line 63: just for curiosity, does it means that the `serviceAccountName` is under the `rbacv1.PolicyRule`?

pkg/knuu/instance_helper.go
```diff
@@ -173,6 +190,20 @@ func (i *Instance) destroyPod() error {
 		return fmt.Errorf("failed to delete pod: %v", err)
 	}
 
+	// Delete the service account for the pod
+	if err := k8s.DeleteServiceAccount(k8s.Namespace(), i.k8sName); err != nil {
```
Bidon15, line 194: Is there a reason to not use the instance's public method of DeleteServiceAccount you defined in `k8s_serviceaccount.go`?

pkg/knuu/instance_helper.go
```diff
@@ -124,6 +124,23 @@ func (i *Instance) deployPod() error {
 		return fmt.Errorf("failed to get image name: %v", err)
 	}
 
+	serviceAccountName := i.k8sName
```
jrmanes, line 127: sa = podName, right?

pkg/knuu/instance_helper.go
```diff
@@ -173,6 +190,20 @@ func (i *Instance) destroyPod() error {
 		return fmt.Errorf("failed to delete pod: %v", err)
 	}
 
+	// Delete the service account for the pod
+	if err := k8s.DeleteServiceAccount(k8s.Namespace(), i.k8sName); err != nil {
+		return fmt.Errorf("failed to delete service account: %v", err)
+	}
+	// Delete the role and role binding for the pod if there are policy rules
+	if len(i.policyRules) > 0 {
+		if err := k8s.DeleteRole(k8s.Namespace(), i.k8sName); err != nil {
```
jrmanes, line 199: not sure (i haven't tested it), but not sure if we'll have an error if we try to remove a role which is being used by the rolebinding

smuu: I added a TODO to the PR. This PR needs more work.

smuu:
This PR needs some more work. I will put it in draft mode until done.

pkg/knuu/instance_helper.go
smuu: I'm calling this function here

pkg/knuu/instance_helper.go
smuu: Reverted back in: https://github.com/celestiaorg/knuu/pull/129/commits/bbda78c6f4d51e1882a5b44d6aebfe73896dbac8

pkg/knuu/instance_helper.go
smuu: > not sure (i haven't tested it), but not sure if we'll have an error if we try to remove a role which is being used by the rolebinding

GPT-4: If you delete a Role being used by a RoleBinding, the RoleBinding will still exist, but it will be ineffective. This is because the RoleBinding references the Role but does not contain the Role's permissions. When the Role is deleted, the permissions it granted are also deleted.

pkg/knuu/instance_helper.go
smuu: Yes, as each Pod will get its own SA, I will use the StatefulSet name as the SA name.

Bidon15 (approved):
🚢

Issue #147, opened 2023-07-11
Add Port Check to Destroy Function to Match Start Function Behavior


smuu:
When calling `Start` the code checks if ports and only deploys a service if so.

https://github.com/celestiaorg/knuu/blob/920d628be85f2670857a962e78481666994feea2/pkg/knuu/instance.go#L619

In the `Destroy`, we don't have this check. We just call the destroy function.

https://github.com/celestiaorg/knuu/blob/920d628be85f2670857a962e78481666994feea2/pkg/knuu/instance.go#L777

Add the same check to the `Destroy` as we have in `Start` for the Kubernetes service.

Frenzoid:
@smuu check this out :) https://github.com/celestiaorg/knuu/pull/149

smuu:
outdated, done differently

PR #149, opened 2023-07-11
Fix: Add Port Check to Destroy Function to Match Start Function Behavior

Frenzoid:
## Overview

Add Port Check to Destroy Function to Match Start Function Behavior
Addressing issue: https://github.com/celestiaorg/knuu/issues/147

imagen

## Checklist

- [ X ] New and updated code has appropriate documentation
- [ X ] New and updated code has new and/or updated testing
- [ X ] Required CI checks are passing
- [ X ] Visual proof for any user facing features like CLI or documentation updates
- [ X ] Linked issues closed with keywords

### DIFF
--- a/pkg/knuu/instance.go
+++ b/pkg/knuu/instance.go
@@ -517,12 +517,10 @@ func (i *Instance) SetEnvironmentVariable(key string, value string) error {
 	logrus.Debugf("Set environment variable '%s' to '%s' in instance '%s'", key, value, i.name)
 	return nil
 }
-
 // GetIP returns the IP of the instance
 // This function can only be called in the states 'Preparing' and 'Started'
 func (i *Instance) GetIP() (string, error) {
-	svc, _ := k8s.GetService(k8s.Namespace(), i.k8sName)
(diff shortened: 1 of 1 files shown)


smuu (changes requested):
I just realized a service can be deployed even if no port was added. See here: https://github.com/celestiaorg/knuu/blob/fd85a081e5c0e9472d0a739b94d8d3807784edb1/pkg/knuu/instance.go#L540

Therefore, I propose to change the condition to `if i.kubernetesService != nil`.
This variable is set when the service is deployed and therefore is `nil` if the service was not deployed.

Frenzoid:
> I just realized a service can be deployed even if no port was added. See here:
> 
> https://github.com/celestiaorg/knuu/blob/fd85a081e5c0e9472d0a739b94d8d3807784edb1/pkg/knuu/instance.go#L540
> 
> Therefore, I propose to change the condition to `if i.kubernetesService != nil`. This variable is set when the service is deployed and therefore is `nil` if the service was not deployed.

Something like this then, right?

imagen

PR #152, opened 2023-07-13
feat: add capability of external dns services

smuu:
Resolves https://github.com/celestiaorg/knuu/issues/120

Tested here: https://github.com/celestiaorg/knuu-example/blob/8439f2ba8104b84464a48ca57fa6c734e05fc6b8/basic/external-dns_test.go

## Checklist

### DIFF
--- a/pkg/k8s/k8s_service.go
+++ b/pkg/k8s/k8s_service.go
@@ -26,9 +26,9 @@ func GetService(namespace, name string) (*v1.Service, error) {
 }
 
 // DeployService deploys a service if it does not exist.
-func DeployService(namespace, name string, labels, selectorMap map[string]string, portsTCP []int, portsUDP []int) (*v1.Service, error) {
+func DeployService(namespace, name string, labels, annotations, selectorMap map[string]string, portsTCP []int, portsUDP []int, serviceType v1.ServiceType) (*v1.Service, error) {
 
-	svc, err := prepareService(namespace, name, labels, selectorMap, portsTCP, portsUDP)
+	svc, err := prepareService(namespace, name, labels, annotations, selectorMap, portsTCP, portsUDP, serviceType)
 	if err != nil {
 		return nil, fmt.Errorf("error preparing service %s: %w", name, err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/k8s/k8s_service.go
```diff
@@ -28,9 +28,9 @@ func GetService(namespace, name string) (*v1.Service, error) {
 }
 
 // DeployService deploys a service if it does not exist.
-func DeployService(namespace, name string, labels, selectorMap map[string]string, portsTCP []int, portsUDP []int) (*v1.Service, error) {
+func DeployService(namespace, name string, labels, annotations, selectorMap map[string]string, portsTCP []int, portsUDP []int, serviceType v1.ServiceType) (*v1.Service, error) {
```
jrmanes, line 31: as this function is growing, what do you think if for a better readability we change the style to:

```go
func DeployService(
	namespace, 
	name string, 
	labels, 
	annotations, 
	selectorMap map[string]string, 
	portsTCP []int, 
	portsUDP []int, 
	serviceType v1.ServiceType) 
(*v1.Service, error) { 
```

pkg/k8s/k8s_service.go
```diff
@@ -126,8 +126,8 @@ func buildPorts(tcpPorts, udpPorts []int) []v1.ServicePort {
 }
 
 // prepareService constructs a new Service object with the specified parameters.
-func prepareService(namespace, name string, labels, selectorMap map[string]string,
-	tcpPorts, udpPorts []int) (*v1.Service, error) {
+func prepareService(namespace, name string, labels, annotations, selectorMap map[string]string,
```
jrmanes, line 129: same to this one, using the previous style

pkg/knuu/instance.go
```diff
@@ -609,6 +610,22 @@ func (i *Instance) SetStartupProbe(startupProbe *v1.Probe) error {
...
+	}
+	// check if external DNS is already added
+	for _, d := range i.externalDns {
+		if d == dns {
+			return fmt.Errorf("external DNS '%s' is already added", dns)
+		}
+	}
+	i.externalDns = append(i.externalDns, dns)
+	return nil
+}
```
Bidon15, line 627: Our of curiosity, is it planned to have removeExternalDNS in the future?

pkg/k8s/k8s_service.go
smuu: https://github.com/celestiaorg/knuu/pull/152/commits/e371faab7c9f662bef4d180d9ad496c4a8a6a60f

pkg/k8s/k8s_service.go
smuu: https://github.com/celestiaorg/knuu/pull/152/commits/e371faab7c9f662bef4d180d9ad496c4a8a6a60f

pkg/knuu/instance.go
smuu: I don't see the use case right now. I could add it for completion tho.

smuu: Agreed with @jrmanes that knuu will not alter an existing environment but rather deploy a new one.

PR #168, opened 2023-07-27
feat: add sidecar functionality

smuu:
## Checklist

### DIFF
--- a/pkg/k8s/k8s_pod.go
+++ b/pkg/k8s/k8s_pod.go
@@ -89,24 +89,31 @@ func NewFile(source, dest string) *File {
 	}
 }
 
+// ContainerConfig contains the specifications for creating a new Container object
+type ContainerConfig struct {
+	Name           string            // Name to assign to the Container
+	Image          string            // Name of the container image to use for the container
+	Command        []string          // Command to run in the container
+	Args           []string          // Arguments to pass to the command in the container
+	Env            map[string]string // Environment variables to set in the container
+	Volumes        []*Volume         // Volumes to mount in the Pod
+	MemoryRequest  string            // Memory request for the container
+	MemoryLimit    string            // Memory limit for the container
+	CPURequest     string            // CPU request for the container
+	LivenessProbe  *v1.Probe         // Liveness probe for the container
+	ReadinessProbe *v1.Probe         // Readiness probe for the container
+	StartupProbe   *v1.Probe         // Startup probe for the container
+	Files          []*File           // Files to add to the Pod
+}
+
 // PodConfig contains the specifications for creating a new Pod object
 type PodConfig struct {
 	Namespace          string            // Kubernetes namespace of the Pod
 	Name               string            // Name to assign to the Pod
 	Labels             map[string]string // Labels to apply to the Pod
-	Image              string            // Name of the Docker image to use for the container
-	Command            []string          // Command to run in the container
-	Args               []string          // Arguments to pass to the command in the container
-	Env                map[string]string // Environment variables to set in the container
-	Volumes            []*Volume         // Volumes to mount in the Pod
-	MemoryRequest      string            // Memory request for the container
-	MemoryLimit        string            // Memory limit for the container
-	CPURequest         string            // CPU request for the container
 	ServiceAccountName string            // ServiceAccount to assign to Pod
-	LivenessProbe      *v1.Probe         // Liveness probe for the container
-	ReadinessProbe     *v1.Probe         // Readiness probe for the container
-	StartupProbe       *v1.Probe         // Startup probe for the container
-	Files              []*File           // Files to add to the Pod
+	ContainerConfig    ContainerConfig   // ContainerConfig for the Pod
+	SidecarConfigs     []ContainerConfig // SideCarConfigs for the Pod
 }
 
 // ReplacePodWithGracePeriod replaces a pod in the given namespace and returns the new Pod object with a grace period.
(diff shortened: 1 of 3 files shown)


pkg/k8s/k8s_pod.go
```diff
@@ -410,82 +417,118 @@ func buildResources(memoryRequest string, memoryLimit string, cpuRequest string)
...
-				Name:  "volume-whatever",
-				Image: image,
-				SecurityContext: &v1.SecurityContext{
-					RunAsUser: &user,
-				},
-				Command:      initContainerCommand,
-				VolumeMounts: initContainerVolumes,
+	return []v1.Container{
+		{
+			Name:  "volume-whatever",
```
jrmanes, line 469: the name is okay or should be:
```suggestion
			Name:  config.Name,
```

pkg/k8s/k8s_networkpolicy.go
```diff
@@ -86,3 +86,29 @@ func DeleteNetworkPolicy(namespace string, name string) error {
...
+
+	return np, nil
+}
+
+// NetworkPolicyExists checks if a NetworkPolicy resource exists.
+func NetworkPolicyExists(namespace string, name string) bool {
+	_, err := GetNetworkPolicy(namespace, name)
+	if err != nil {
+		return false
+	}
```
Bidon15, line 111: [non-blocking]: I would put some debug logs here on what the error msg was

pkg/knuu/instance.go
```diff
@@ -285,28 +236,46 @@ func (i *Instance) ExecuteCommand(command ...string) (string, error) {
...
 		commandWithShell := []string{"/bin/sh", "-c", strings.Join(command, " ")}
-		output, err := k8s.RunCommandInPod(k8s.Namespace(), pod.Name, i.k8sName, commandWithShell)
+		output, err := k8s.RunCommandInPod(k8s.Namespace(), pod.Name, containerName, commandWithShell)
 		if err != nil {
-			return "", fmt.Errorf("error executing command '%s' in started instance '%s': %v", command, i.k8sName, err)
+			if i.isSidecar {
+				return "", fmt.Errorf("error executing command '%s' in sidecar '%s' of instance '%s': %v", command, i.k8sName, i.parentInstance.k8sName, err)
+			} else {
+				return "", fmt.Errorf("error executing command '%s' in instance '%s': %v", command, i.k8sName, err)
+			}
```
Bidon15, line 256: I would make the Sidecar checker as a separate if statement outside of the parent container checker. 
Have 3 nested ifs is not readable.

pkg/knuu/instance.go
```diff
@@ -626,49 +595,83 @@ func (i *Instance) SetStartupProbe(startupProbe *v1.Probe) error {
...
+			return fmt.Errorf("error deploying resources for instance '%s': %w", i.k8sName, err)
 		}
-		if len(i.files) != 0 {
-			err := i.deployFiles()
-			if err != nil {
-				return fmt.Errorf("error deploying files for instance '%s': %w", i.k8sName, err)
+		for _, sidecar := range i.sidecars {
+			if err := sidecar.deployResources(); err != nil {
+				return fmt.Errorf("error deploying resources for sidecar '%s': %w", sidecar.k8sName, err)
 			}
```
Bidon15, line 650: you can refactor the whole thing into a func that accepts a functions as a signature and ranges the given sidecar variable. This will eliminate most of the ranges and keep the code cleaner

pkg/knuu/instance_helper.go
```diff
@@ -274,8 +243,81 @@ func (i *Instance) destroyFiles() error {
...
+			if err != nil {
+				return fmt.Errorf("error deploying service for instance '%s': %w", i.k8sName, err)
+			}
+		} else if svc != nil {
+			err := i.patchService()
+			if err != nil {
+				return fmt.Errorf("error patching service for instance '%s': %w", i.k8sName, err)
+			}
+		}
+	}
```
Bidon15, line 262: this should be refactored into separate private functions

Bidon15 (changes requested):

pkg/k8s/k8s_networkpolicy.go
smuu: https://github.com/celestiaorg/knuu/pull/168/commits/ad4a8e6024de48d1f4b4ae0450f5b9377e345784

pkg/knuu/instance.go
smuu: Do you mean implement a function `handleCommandError(err, command, i)`?

Another option is not obvious to me, as I want to return an error depending on if the instance the command execution threw an error is a sidecar or not.
But to simplify it, we could use the same error message for both.

pkg/knuu/instance_helper.go
smuu: https://github.com/celestiaorg/knuu/pull/168/commits/0326589d701f1321c388d9056049a6a621fc9017

pkg/knuu/instance.go
Bidon15: Yes! Fantastic. 
Still, I'd name it differently from performActions

pkg/knuu/instance.go
Bidon15: Having a func is another solution that sgtm.

You can check if the instance is the sidecar or instance first before assigning the error. The idea is to as explicit as possible due to the size of the struct. In addition to that, reducing the amount of nested statements 

For example: 
```go
output, err := k8s.RunCommandInPod(k8s.Namespace(), pod.Name, containerName, commandWithShell)
if i.isSidecar && err != nil {
    return "", fmt.Errorf("error executing command '%s' in sidecar '%s' of instance '%s': %v", command, i.k8sName, i.parentInstance.k8sName, err)
} else if !i.isSidecar && err != nil{
    return "", fmt.Errorf("error executing command '%s' in instance '%s': %v", command, i.k8sName, err)
}
```

Another can be regarding the whole else if statement where you first check for the isSidecar and then have the general logic
```go
var instanceName string
var errMsg err
containerName := i.k8sName

if isSidecar {
  instanceName = i.parentInstance.k8sName
  errMsg = fmt.Errorf("error executing command '%s' in sidecar '%s' of instance '%s': %v", command, i.k8sName, i.parentInstance.k8sName, err)
} else {
  // whatever opposite for instance
}
output, err := k8s.RunCommandInPod(k8s.Namespace(), pod.Name, containerName, commandWithShell)
if err != nil {
   return "", errMsg
}

return "", nil

```

smuu: Thank you for the feedback! See https://github.com/celestiaorg/knuu/pull/168/commits/49bd378404e5111e1907835989e76da4bf1a94f5 and https://github.com/celestiaorg/knuu/pull/168/commits/a57b615786fc8180d6a2bf14fc9bd3c5cd50a681.

Now the code is much more readable :+1:

pkg/knuu/instance.go
smuu: Renamed and refactored the function to be more generic: https://github.com/celestiaorg/knuu/pull/168/commits/9151c3e8f646e0fc661166cbcdbf030712de5209

smuu:
Thank you @Bidon15 for the great review!

PR #220, opened 2023-12-14
chore: first draft of the ADR 001

### LINKED ISSUE #215: Test instances orchestration
## Problem

Currently if we want to shape the traffic with `Knuu+BitTwsiter`, we need to expose the address & ports of the pod running the target instance with a sidecar running BitTwister to be able to access to BitTwister from Knuu. The main reason is because Knuu is running from above and creates instances, configures them and runs them.

### Reverse proxy
@smuu please fill this section with the idea

### Distributed Playbook

Since most (if not all) tests have plans before hand, it is known when and how to shape the traffic. The playbook idea is a way similar to what testground is doing, and it is the following:

1. User defines the test scenario using e.g. golang or some other descriptive language can be just a JSON file
2. Knuu initiates the instances and assigns each playbook to their corresponding instance
3. BitTwister is running in a sidecar with some extra code that understand the playbook
4. It changes the traffic shape at the time that is required. it can be any given event from the target instance.

**Advantages:** 

1. No need for a centralized entity to instruct each instance what to do. Everyone knows what to do beforehand just a like a bee.
2. No need to expose IP addresses or using a proxy for the `control data`

mojtaba-esk:
This PR compiles what have been discussed on redesigning knuu with an async approach.

Please keep the discussions here to hammer it and come up with a high level design roadmap for `Knuu v1`

Closes #215


docs/architecture/adr-001-redesign-knuu-async-approach.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Proposed Solution
+
+### A Distributed Approach with a Playbook Model
+
+Embrace a distributed playbook model similar to Testground, allowing asynchronous orchestration and execution of test scenarios:
+
+1. **Test Scenario Definition:** Users define test scenarios in a descriptive language (e.g., Golang).
+2. **Knuu Instance Initiation:** Knuu initiates instances and assigns specific playbooks to each instance.
+3. **BitTwister Sidecar with Playbook Understanding:** BitTwister runs alongside instances with added functionality to interpret and execute the playbook's instructions.
```
Bidon15, line 23: As @smuu stated in one of our syncs, we can have a scenario, where we have 1 k8s pod containing 3-4-5-6-N containers per pod. Which means we can technically have 200k~ containers while using 100 pods. 

Meaning that each of 100 pod can contain 1 BitTwister sidecar and 1800 containers, each test container will have 1 network condition behaviour. 

Thus, we can have 100k requirement achieved internally for node team

docs/architecture/adr-001-redesign-knuu-async-approach.md
```diff
@@ -0,0 +1,49 @@
...
+4. **Dynamic Traffic Shaping:** BitTwister adjusts traffic shaping based on predefined events or triggers from the target instance, eliminating the need for continuous centralized control.
+
+### Advantages
+
+1. **Decentralized Execution:** Instances operate autonomously based on pre-defined playbooks, reducing dependency on centralized control.
+2. **Elimination of IP Exposure:** Removes the necessity to expose IP addresses or utilize proxies for controlling data.
```
Bidon15, line 29: Private IPs is the king of the game here. I would add that all test containers need only private IPs to talk to one another in a cluster. 

When we do multiclusters - we will figure that out -> this point should be pointed in Considerations / Risks point

docs/architecture/adr-001-redesign-knuu-async-approach.md
```diff
@@ -0,0 +1,49 @@
...
+
+The current synchronous model necessitates exposing the target instance's address and ports to BitTwister via a sidecar. This centralized approach introduces dependencies and complexities in configuring the test environment. Additionally, the reliance on a proxy approach or port forwarding poses scalability issues, resulting in inconsistent errors during test execution.
+
+## Proposed Solution
+
+### A Distributed Approach with a Playbook Model
+
+Embrace a distributed playbook model similar to Testground, allowing asynchronous orchestration and execution of test scenarios:
+
+1. **Test Scenario Definition:** Users define test scenarios in a descriptive language (e.g., Golang).
```
Bidon15, line 21: CI/CD gateway is needed here as most of tests should be cron jobs for verification pipelines for upcoming releases of artifacts owned by the contributors, etc.

Bidon15 (changes requested):
Awesome job! Left some comments

docs/architecture/adr-001-redesign-knuu-async-approach.md
smuu, line 23: I would propose decoupling the playbook logic from bittwister, as this playbook can have other use cases such as injecting transactions, restarting the node, etc.

smuu (changes requested):

docs/architecture/adr-001-redesign-knuu-async-approach.md
```diff
@@ -0,0 +1,49 @@
...
+2. **Elimination of IP Exposure:** Removes the necessity to expose IP addresses or utilize proxies for controlling data.
+
+### Comments
+
+Further discussions revealed the necessity for test designers to predefine scenarios and program them before runtime, a feature lacking in the current synchronous design of Knuu. Moving towards an asynchronous-only approach aligns with the distributed design of Testground and simplifies usage.
+
+### Proposed Roadmap
+
+- **v0:** Consider the current version as complete, encapsulating UX issues with helper functions where needed.
+- **v1:** Create a separate branch to implement a new ADR-driven design. This version will gather consensus on the foundational aspects of the new design before delving into implementation details.
```
ramin, line 38: yes

docs/architecture/adr-001-redesign-knuu-async-approach.md
ramin: i'd propose looking at something cue for conifg https://cuelang.org/, a bit like HCL but with a bit more dynamism, dagger uses it for it's pipelines and it looks quite nice

ramin (commented):
some comments! would love to join some of these discussions if welcome

smuu:
Let me know what you think about the following idea:

The Bee Playbook is configured via a YAML. Below, I have for you an example with packet loss. I tried to have the format of the YAML as generic as possible, so that we can extend it.
Then, knuu could provide a wrapper of that YAML in the Go code so that the developer can specify the behavior with code instead of fighting with YAML.

```yaml
endpoints:
- name: "rpc"
  endpoint: "0.0.0.0:26657"
- name: "bittwister"
  endpoint: "0.0.0.0:9009"
sources:
- name: "status"
  endpoint: "rpc"
  path: "/status"
  polling-time: "1s"
targets:
- name: "packetloss"
  endpoint: "bittwister"
  path: "/packetloss"
  values:
  - "rate"
parser:
- name: "height"
  type: "json"
  command: ".result.sync_info.latest_block_height"
playbook:
- source: "status"
  parser: "height"
  conditions:
  - condition: ">="
    value: 100
  - condition: "<="
    value: 200
  target:
    name: "packetloss"
    values:
      rate: 25
```

docs/architecture/adr-001-redesign-knuu-async-approach.md
mojtaba-esk: Good ideas, thanks guys. To have things clean and reduce communication overhead, we can embed BitTwister code inside that playbook. It can simply import it and control it according to the given plan. I do agree to keep its code base separate and allow users to plan their traffic shape scenario as they want using for example a go code that makes the playbook at the end.

MSevey (commented):
will review once initial comments are addressed.

docs/architecture/adr-001-redesign-knuu-async-approach.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Problem
+
+The current synchronous model necessitates exposing the target instance's address and ports to BitTwister via a sidecar. This centralized approach introduces dependencies and complexities in configuring the test environment. Additionally, the reliance on a proxy approach or port forwarding poses scalability issues, resulting in inconsistent errors during test execution.
+
+## Proposed Solution
+
+### A Distributed Approach with a Playbook Model
+
+Embrace a distributed playbook model similar to Testground, allowing asynchronous orchestration and execution of test scenarios:
```
MSevey, line 19: What will be the differences in use case between KNUU and Testground if we move in this direction? 

I think something that would be helpful for this ADR, and any large architecture overhaul, is to include some examples of desired test scenarios that we want to be able to support.

MSevey (changes requested):
One small comment on some example test cases we expect to, or want to, be able to support. 
I think this would be helpful in highlighting any other considerations we should add to the ADR. 

otherwise lgtm 👍

docs/architecture/adr-001-redesign-knuu-async-approach.md
mojtaba-esk: Well in that case Knuu will be going towards the same direction of TG and we have to see which one to choose to work on. 
Knuu is built in house and needs a lot of love.
TG seems to be used by more than one team.

That's right we need to have some good examples. @Bidon15 can you please shed some lights here?

mojtaba-esk: Ok here is an example that I could come up with:

Let's say we have the following node Configuration:

- **150** Validators
- **300** Consensus Full Nodes
- **1000** Full Data Availability (DA) Nodes
- **100,000** Light DA Nodes

Playbook setup:
- **After reaching height 10**:
    - Adjust the bandwidth of 15% of Validators to 10Mbps.
    - Introduce a 50ms latency to 30% of Light DA Nodes.
    - Add a 20ms jitter to 25% of Full DA Nodes.

- **Test Duration**:
    - Keep the test running until height 10,000 
    - or 30 minutes, ensuring comprehensive evaluation.

- **Metrics collection**
    - Configured traffic shape on each node
    - Error logs of each node
    - Time to complete the sync of DA nodes
    - Number of missing blocks to sign
    - Number of missing blocks to propose
    - ...


Each worker node receives this playbook and based on that decides what to do.
The entire playbook could be written in a golang code just like TG.

ramin (commented):
I think a paragraph on how the asynchronous approach might be achieved and some of the primitives we envision having to both build, and that which might be a challenge, would be at least worth mentioning: clock sync / block time ? drift? consistency in decentralized plan state changes (eg: test run 1 - 30% of nodes targeted with a +100ms latency injection enabled within 10 seconds, test run 2 - 10% failed to transition for whatever reason) etc. 

Lets outline the challenges and rabbit holes is my point I guess

docs/architecture/adr-001-redesign-knuu-async-approach.md
mojtaba-esk: if we go either way of improving Knuu or work on TG, for both we will need heavy refactor. or rather building a significant portion of it. Each of them has it's own potentials and it is a bit hard to chose either ways. The structure that is proposed in this ADR is more align with TG; however, TG has a number of significant limitation that need to be resolved to be able to handle a large scale test.

@Bidon15 has worked a lot with TG, tell us what do you think?

MSevey: ah ok makes sense. 
So we could have a future where we just use knuu and testground is dead to us. 

Makes sense. 

Let's add the example that you listed here to the ADR as a reference.

smuu:
Let's merge this while keeping the ADR in draft. If we have more input, let's create issues for it.

mojtaba-esk:
> Let's merge this while keeping the ADR in draft. If we have more input, let's create issues for it.

Good idea. Let's do it. Right now it cannot be merged as it says: `Merging can be performed automatically once the requested changes are addressed` @ramin and @MSevey must approve it first as they requested some changes.

MSevey (approved):

mojtaba-esk:
> > Let's merge this while keeping the ADR in draft. If we have more input, let's create issues for it.
> 
> Good idea. Let's do it. Right now it cannot be merged as it says: `Merging can be performed automatically once the requested changes are addressed` @Bidon15 @ramin and @MSevey must approve it first as they requested some changes.

@Bidon15 and @smuu  can you please approve this one?

smuu (approved):

PR #227, opened 2024-01-17
feat: implement kaniko builder

mojtaba-esk:
This PR proposes an implementation for a Builder interface and a Kaniko builder implementation for it.
It attempts to create a k8s job, build the given image from a github repo url (or with a commit hash), then pushes it to a registry. It keeps track of the building process and returns the logs at the end.

Please note that, it has not yet been integrated into Knuu. To keep the review process easier, it will be handled in a separate PR.

### DIFF
--- a/pkg/builder/kaniko/git.go
+++ b/pkg/builder/kaniko/git.go
@@ -5,6 +5,12 @@ import (
 	"strings"
 )
 
+const (
+	regexpGitRepoProtocol = `^(https?|git|ssh|ftp)://`
+	regexpGitRepoDotGit   = `\.git$`
+	gitProtocol           = "git://"
+)
+
 type GitContext struct {
 	Repo     string
 	Commit   string
@@ -12,24 +18,24 @@ type GitContext struct {
 	Password string
 }
 
-func GetGitBuildContext(g *GitContext) (string, error) {
+func (g *GitContext) BuildContext() (string, error) {
 	bCtx := ""
 
 	// cleaning the repo url
-	rgx, err := regexp.Compile(`^(https?|git|ssh|ftp)://`)
+	rgx, err := regexp.Compile(regexpGitRepoProtocol)
 	if err != nil {
 		return "", err
 	}
 	g.Repo = rgx.ReplaceAllString(g.Repo, "")
 
-	rgx, err = regexp.Compile(`\.git$`)
+	rgx, err = regexp.Compile(regexpGitRepoDotGit)
 	if err != nil {
 		return "", err
 	}
 	g.Repo = rgx.ReplaceAllString(g.Repo, "")
 	g.Repo = strings.TrimSuffix(g.Repo, "/")
 
-	bCtx += "git://"
+	bCtx += gitProtocol
 	if g.Username != "" {
 		bCtx += g.Username
 		if g.Password != "" {
--- a/pkg/builder/kaniko/git.go
+++ b/pkg/builder/kaniko/git.go
@@ -1,0 +1,47 @@
+package kaniko
+
+import (
+	"regexp"
+	"strings"
+)
+
+type GitContext struct {
+	Repo     string
+	Commit   string
+	Username string
+	Password string
+}
+
+func GetGitBuildContext(g *GitContext) (string, error) {
+	bCtx := ""
+
+	// cleaning the repo url
+	rgx, err := regexp.Compile(`^(https?|git|ssh|ftp)://`)
+	if err != nil {
+		return "", err
+	}
+	g.Repo = rgx.ReplaceAllString(g.Repo, "")
+
+	rgx, err = regexp.Compile(`\.git$`)
+	if err != nil {
+		return "", err
+	}
+	g.Repo = rgx.ReplaceAllString(g.Repo, "")
+	g.Repo = strings.TrimSuffix(g.Repo, "/")
+
+	bCtx += "git://"
+	if g.Username != "" {
+		bCtx += g.Username
+		if g.Password != "" {
+			bCtx += ":" + g.Password
+		}
+		bCtx += "@"
+	}
+
+	bCtx += g.Repo
+	if g.Commit != "" {
+		bCtx += "#" + g.Commit
+	}
+
+	return bCtx, nil
+}
--- a/pkg/builder/kaniko/kaniko.go
+++ b/pkg/builder/kaniko/kaniko.go
@@ -1,0 +1,248 @@
+package kaniko
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+
+	"github.com/celestiaorg/knuu/pkg/builder"
+	"github.com/google/uuid"
+	batchv1 "k8s.io/api/batch/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	kanikoImage         = "gcr.io/kaniko-project/executor:debug" // debug has a shell
+	kanikoContainerName = "kaniko-container"
+	kanikoJobNamePrefix = "kaniko-build-job"
+)
+
+type Kaniko struct {
+	k8sClientset kubernetes.Interface
+	k8sNamespace string
+}
+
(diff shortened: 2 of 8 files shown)


pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+								`--context=` + b.BuildContext,
+								// TODO: see if we need it or not
+								// --git gitoptions    Branch to clone if build context is a git repository (default branch=,single-branch=false,recurse-submodules=false)
+
+								// TODO: we might need to add some options to get the auth token for the registry
+								"--destination=" + b.Destination,
+							},
+						},
+					},
+					RestartPolicy: "IfFailed",
```
mojtaba-esk, line 217: Not sure if the `RestartPolicy` is set correctly. Right now if there is something wrong with the `Dockerfile` it fails and keeps trying so it is assumed that there is nothing wrong with the target application build process.
We might want to change this. Please share your thoughts on it

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	if len(podList.Items) == 0 {
+		return nil, fmt.Errorf("no Pods found for the Job: %s", job.Name)
+	}
+
+	return &podList.Items[0], nil
+}
+
+func (k *Kaniko) getContainerLogs(ctx context.Context, pod *v1.Pod) (string, error) {
```
ramin, line 129: I'd remove all the "gets" from these personally, will read more beautifully to be k.ContainerLogs() etc

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+type Kaniko struct {
+	k8sClientset kubernetes.Interface
+	k8sNamespace string
+}
+
+var _ builder.Builder = &Kaniko{}
+
+func NewKaniko(k8sClientset kubernetes.Interface, namespace string) *Kaniko {
```
ramin, line 30: because we're already inside kaniko package, I'd rename this New, so it reads `kaniko.New` vs `kaniko.NewKaniko`

potentially would be more idiomatic to pass in an options struct and have it just take one argument with the k8sClientset and namespace in the options IMO

pkg/builder/kaniko/git.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type GitContext struct {
+	Repo     string
+	Commit   string
+	Username string
+	Password string
+}
+
+func GetGitBuildContext(g *GitContext) (string, error) {
```
ramin, line 15: feel like this should be defined on GitContext, vs passed in, so maybe it reads more like

```go

gc := GitContext{...} // fill with stuff
gc.AsContextString() 
// OR
gc.BuildContext() 
```

etc, would look nice...

pkg/builder/kaniko/git.go
```diff
@@ -0,0 +1,47 @@
...
+	bCtx := ""
+
+	// cleaning the repo url
+	rgx, err := regexp.Compile(`^(https?|git|ssh|ftp)://`)
```
ramin, line 19: I'd move the regexp string up to a const then make it easy to grab just the regexp and test/access

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+
+			if j.Status.Succeeded > 0 || j.Status.Failed > 0 {
+				// Job completed (successfully or failed)
+				return j, nil
+			}
+		case <-ctx.Done():
+			return nil, fmt.Errorf("context cancelled")
+		}
+	}
+}
```
ramin, line 111: love this

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+			LabelSelector: fmt.Sprintf("job-name=%s", job.Name),
+		})
+	if err != nil {
+		return fmt.Errorf("error deleting Pods: %w", err)
+	}
+
+	return nil
+}
+
+func GetDefaultCacheOptions(buildContext string) (*builder.CacheOptions, error) {
```
ramin, line 170: again, DefaultCacheOptions IMO

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+	if buildContext == "" {
+		return nil, fmt.Errorf("build context cannot be empty")
```
ramin, line 172: I'm a fan of potentially defining an `errors.go` file and naming these kind of errors, so they can be explicitly referenced/inspected outside, especially if there are multiple implementations of the builder, ie:

builder/errors#ErrEmpyBuildContext

then each implementation and return similar errors

pkg/builder/kaniko/kaniko.go
ramin: should this be OnFailure? If you want it to die, it should be `Never` I believe, what is desired behavior?

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	// Add extra args
+	job.Spec.Template.Spec.Containers[0].Args = append(job.Spec.Template.Spec.Containers[0].Args, b.Args...)
+
+	return job, nil
+
+}
+
+func generateK8sName(name string) (string, error) {
```
ramin, line 242: feels like generating pod / resource names should maybe live on its own in a package? eg: `names.NewRandom()`

ramin (commented):
pretty great

pkg/builder/kaniko/git.go
mojtaba-esk: good point, thanks :)

pkg/builder/kaniko/kaniko.go
mojtaba-esk: delicate point, thanks. will apply it asap

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,248 @@
...
+	_, err := hash.Write([]byte(buildContext))
+	if err != nil {
+		return nil, err
+	}
+	hashStr := hex.EncodeToString(hash.Sum(nil))
+
+	return &builder.CacheOptions{
+		Enabled: true,
+		Dir:     "",
+		Repo:    fmt.Sprintf("ttl.sh/%s:24h", hashStr),
```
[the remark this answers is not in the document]
mojtaba-esk: This is just a default cache option and user can decide to use it or define whatever they want. and yes user must define destination when building an image.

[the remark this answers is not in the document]
mojtaba-esk: we can have some default destination as well

pkg/builder/kaniko/kaniko.go
mojtaba-esk: I think we can leave it to try a couple of times then fail completely.

mojtaba-esk: @jrmanes it seems like the default value for `BackoffLimit` is `6` can I just use that instead of introducing the index thing?

pkg/builder/kaniko/kaniko.go
mojtaba-esk: Good idea we can do that and use it across the whole repo

pkg/builder/kaniko/kaniko.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func prepareJob(b *builder.BuilderOptions) (*batchv1.Job, error) {
+	jobName, err := names.NewRandomK8(kanikoJobNamePrefix)
+	if err != nil {
+		return nil, ErrGeneratingUUID.Wrap(err)
+	}
+
+	oneInt32 := int32(1)
+	fiveInt32 := int32(5)
```
ramin, line 175: any reason not to set these at top of the package with proper name, ie:

```go
var (
   DefaultParallelism = int32(5)
   BackoffLimit = int32(5)
)
```

ramin (approved):
:shipit:

pkg/builder/kaniko/kaniko.go
mojtaba-esk: nice catch, will fix it in a minute

PR #238, opened 2024-02-01
fix: Mount dir to kaniko using Minio

mojtaba-esk:
Currently, building from a local directory using `kaniko` builder is only possible when the directory is less than 3MB. The main reason is because we use config maps and try to make an archive of the current directory and set it in the Job instructions and since k8s does not allow a request more than 3MB, it fails for most of the repos.

So this PR proposes a new change where it first create a Minio file sharing service in the cluster and then it makes a `tar.gz` archive of the directory and send it to Minio. 
Next, it downloads it from an `init` container into a volume, which is also mounted onto the `Kaniko` container. 

This way we do not need to modify the Kaniko container and maintain another code base, and we can build a local directory on k8s and it is seamless to the user.


pkg/minio/minio.go
```diff
@@ -0,0 +1,418 @@
...
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"app": "minio"}},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{{
+						Name:  DeploymentName,
+						Image: Image,
+						Env: []v1.EnvVar{
+							{Name: "MINIO_ACCESS_KEY", Value: accessKey},
+							{Name: "MINIO_SECRET_KEY", Value: secretKey},
```
Bidon15, line 79: Should be consts

pkg/minio/minio.go
```diff
@@ -0,0 +1,418 @@
...
+
+	// Create Minio deployment
+	minioDeployment := &appsV1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      DeploymentName,
+			Namespace: m.Namespace,
+		},
+		Spec: appsV1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": "minio"},
```
Bidon15, line 69: should be consts

pkg/minio/minio.go
```diff
@@ -0,0 +1,418 @@
...
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"app": "minio"}},
```
Bidon15, line 72: same

pkg/minio/minio.go
```diff
@@ -0,0 +1,418 @@
...
+			existingPV = &pv
+			break
+		}
+	}
+
+	if existingPV == nil {
+		// Create a simple PV if no existing PV is suitable
+		_, err = m.Clientset.CoreV1().PersistentVolumes().Create(ctx, &v1.PersistentVolume{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "minio-pv-",
```
Bidon15, line 374: should be a const prefix

pkg/minio/minio.go
```diff
@@ -0,0 +1,418 @@
...
+			AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce},
+			Resources: v1.VolumeResourceRequirements{
+				Requests: v1.ResourceList{
+					v1.ResourceStorage: storageQt,
+				},
+			},
+		},
+	}
+
+	_, err = pvcClient.Create(ctx, pvc, metav1.CreateOptions{})
```
Bidon15, line 411: why are we not allowing options in signature func?

Bidon15: Ideally, we need to refactor this in the future to accomodate such use-cases

Bidon15 (changes requested):

pkg/minio/minio.go
mojtaba-esk: sure

pkg/minio/minio.go
mojtaba-esk: ok

pkg/minio/minio.go
mojtaba-esk: done

Bidon15 (approved):

PR #239, opened 2024-02-01
fix: builder error propagation

mojtaba-esk:


pkg/builder/kaniko/kaniko.go
```diff
@@ -21,7 +21,7 @@ const (
 	kanikoContainerName = "kaniko-container"
 	kanikoJobNamePrefix = "kaniko-build-job"
 
-	DefaultParallelism  = int32(5)
+	DefaultParallelism  = int32(1)
```
mojtaba-esk, line 24: Needed to set this to one, so the builder pod will be only one as it does not make sense to have more pods for a build task.

ramin (approved):

ramin (approved):

sysrex (approved):
lgtm

PR #277, opened 2024-04-18
fix(scope): use SanitizeName to fix the name of the id

tty47:
hello!

I found a bug in the scope, it was adding the labels with the wrong name, I'm using the `SanitizeName` to fix them, and now they look okay.

Cheers! 🚀


mojtaba-esk (approved):
Awesome

PR #307, opened 2024-04-30
feat(example): env to json

tty47:
hello!

this test adds the content of the `.env` file as json into the container.

cheers! 🚀


mojtaba-esk (approved):
Well done.
Just a question, is passing envs as JSON a common practice? I mean how applications read it and how it translate it to ENV vars?

pkg/knuu/knuu.go
```diff
@@ -36,6 +37,11 @@ var (
 
 // Initialize initializes knuu with a unique scope
 func Initialize() error {
+	err := godotenv.Load()
+	if err != nil {
+		return ErrCannotLoadEnv.Wrap(err)
```
mojtaba-esk, line 42: [question] 
- Does the `.env` file have to exist for all tests?
- if it does not exist will it through an error here?

pkg/knuu/knuu.go
```diff
@@ -36,6 +37,11 @@ var (
...
+	}
```
smuu, line 43: Please move this to `InitializeWithScope()` so it's always loaded.

smuu (changes requested):

.env.example
```diff
@@ -0,0 +1,3 @@
+KNUU_TIMEOUT=60m # The timeout for the tests
+KNUU_BUILDER=docker # The builder to use for building images. docker or kubernetes
+DEBUG_LEVEL=info # The debug level. debug, info, warn, error
```
smuu, line 3: I think this is called `LOG_LEVEL`

```suggestion
LOG_LEVEL=info # The debug level. debug, info, warn, error
```

.env.example
smuu, line 1: Having this `.env` in this repo with good values would be nice when we run the tests.

smuu (changes requested):
Nice one, only small things

smuu (approved):
Let's go! :rocket:

PR #313, opened 2024-05-02
fix(configmap): command fixed to grant permissions to the parent folders

tty47:
hello!

this PR fixes the issue that app team is facing here: https://github.com/celestiaorg/celestia-app/pull/3356

example:
```bash
cp: can't stat '/home/celestia/.celestia-app/*': No such file or directory
```

this code creates the files folders and grants permissions even if they come from a ConfigMap.

cheers! 🚀

### DIFF
--- a/e2e/basic/assert_cleanups.go
+++ b/e2e/basic/assert_cleanups.go
@@ -24,7 +24,6 @@ func assertCleanupInstances(t *testing.T, executor *knuu.Executor, instances []*
 		err := executor.Destroy()
 		if err != nil {
 			t.Fatalf("Error destroying executor: %v", err)
-			return err
 		}
 
 		for _, instance := range instances {


smuu (changes requested):
Nice job!

Please add tests for these cases:

- no volumes and no file
- 1 volume and no file
- no volume and 1 file
- 2 volumes and no file
- no volume and 2 files
- 2 volumes and 2 files

Also, please add the initContainerCommands from these cases to this PR. Maybe we can see how we can improve it to make it better maintainable.

pkg/k8s/k8s_pod.go
```diff
@@ -417,18 +418,38 @@ func buildInitContainerVolumes(name string, volumes []*Volume) ([]v1.VolumeMount
...
+	var commands = []string{"sh", "-c"}
+	dirsProcessed := make(map[string]bool)
+	baseCmd := "mkdir -p /knuu && "
+	cmds := []string{baseCmd}
+
+	for _, file := range files {
+		// get the directory of the file
+		folder := filepath.Dir(file.Dest)
+		if _, processed := dirsProcessed[folder]; !processed {
+			parentDirCmd := fmt.Sprintf("mkdir -p /knuu%s && chmod -R 777 /knuu/%s && ", folder, folder)
```
mojtaba-esk, line 431: Can you please double check if we really need that `/` before `%s` or not?
It seems like there is no `/` in `mkdir` and there is one in `chmod`
we can also use [path.Join](https://pkg.go.dev/path#Join)

pkg/k8s/k8s_pod.go
```diff
@@ -417,18 +418,38 @@ func buildInitContainerVolumes(name string, volumes []*Volume) ([]v1.VolumeMount
...
+			dirsProcessed[folder] = true
+		}
 	}
 
-	var command = []string{"sh", "-c"} // initialize the command slice with the required shell interpreter
-	for _, volume := range volumes {
-		cmd := fmt.Sprintf("mkdir -p /knuu/%s && cp -r %s/* /knuu/%s && chown -R %d:%d /knuu/*", volume.Path, volume.Path, volume.Path, volume.Owner, volume.Owner)
-		command = append(command, cmd) // add each command to the command slice
+	for i, volume := range volumes {
+		cmd := fmt.Sprintf("if [ -d %s ] && [ \"$(ls -A %s)\" ]; then cp -r %s/* /knuu/%s && chown -R %d:%d /knuu/*", volume.Path, volume.Path, volume.Path, volume.Path, volume.Owner, volume.Owner)
```
mojtaba-esk, line 438: [nit] it would be also nice to extract `/knuu` to a constant

mojtaba-esk (dismissed):
Well done :rocket: 
Just left a question and a comment.

smuu (approved):
You :rock:

smuu:
Regarding https://github.com/celestiaorg/knuu/issues/317

Could we add a check in this PR to fail if the dev wants to add more than one volume?

pkg/knuu/instance.go
```diff
@@ -662,6 +662,11 @@ func (i *Instance) AddVolumeWithOwner(path, size string, owner int64) error {
 	if !i.IsInState(Preparing, Committed) {
 		return ErrAddingVolumeNotAllowed.WithParams(i.state.String())
 	}
+	// temporary feat, we will remove it once we can add multiple volumes
+	if len(i.volumes) > 0 {
```
smuu, line 666: Would it make sense to add the check directly here?
https://github.com/celestiaorg/knuu/blob/1012d393b97b889a403ad6f5702b7bae50957277/pkg/knuu/instance.go#L654

smuu: I'm sorry. You where right..

smuu: I'm confused :D Friday mode..

e2e/basic/assert_cleanups.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+// assertCleanupInstance is a helper function that cleans up a single instance.
+func assertCleanupInstance(t *testing.T, instance *knuu.Instance) error {
```
mojtaba-esk, line 11: A nice one. 

[optional] we can add something more generic batch destroy function to the knuu pkg that receives a series of instances e.g:
```go
func BatchDestroy(instance ...*Instance) error {
...
}
```

it might help to have a cleaner test code.

e2e/basic/assert_cleanups.go
```diff
@@ -0,0 +1,42 @@
...
+	return nil
+}
+
+// assertCleanupInstances is a helper function that cleans up a list of instances.
+func assertCleanupInstances(t *testing.T, executor *knuu.Executor, instances []*knuu.Instance) error {
+	if os.Getenv("KNUU_SKIP_CLEANUP") != "true" {
+		err := executor.Destroy()
+		if err != nil {
+			t.Fatalf("Error destroying executor: %v", err)
+			return err
```
mojtaba-esk, line 28: when we use `t.Fatalf` the execution ends so that `return err` is unreachable. It is recommended either to return an error or use Fatalf. Since the name is assert, `Fatalf` might make more sense.

e2e/basic/file_test_image_cached_test.go
```diff
@@ -63,20 +63,9 @@ func TestFileCached(t *testing.T) {
...
-
-			for _, instance := range instances {
-				if instance != nil {
-					err := instance.Destroy()
-					if err != nil {
-						t.Fatalf("Error destroying instance: %v", err)
-					}
-				}
-			}
+		err := assertCleanupInstances(t, executor, instances)
```
mojtaba-esk, line 66: Nice, I like clean cleanup funcs :)

mojtaba-esk (changes requested):
Great work! :+1:

e2e/basic/assert_cleanups.go
```diff
@@ -0,0 +1,53 @@
...
```
mojtaba-esk, line 27: All errors are in a good shape. Only this one is missed out.

mojtaba-esk (changes requested):
just a tiny thing

mojtaba-esk (approved):
Well done :rocket:

smuu (approved):
You :rock:

PR #314, opened 2024-05-02
feat: add build from git example test

### LINKED ISSUE #311: Build from a git repo instead of current local directory
This option is available for the Kaniko builder only and since we are deprecating the docker builder, we can enable this option even now.

mojtaba-esk:
Closes #311

### DIFF
--- a/e2e/basic/build_from_git_test.go
+++ b/e2e/basic/build_from_git_test.go
@@ -1,0 +1,61 @@
+package basic
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/knuu/pkg/builder"
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+// This test is just an example to show how to
+// setup the test instance to be built from a git repo
+func TestBuildFromGit(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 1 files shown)


e2e/basic/build_from_git_test.go
```diff
@@ -0,0 +1,61 @@
...
+	err = instance.SetGitRepo(ctx, builder.GitContext{
+		Repo:     "https://github.com/celestiaorg/celestia-app.git",
+		Branch:   "main",
+		Commit:   "",
+		Username: "",
+		Password: "",
+	})
+	require.NoError(t, err, "Error setting git repo")
+
+	require.NoError(t, instance.Commit(), "Error committing instance")
```
smuu, line 43: Could you show in the test that you can make modifications (like `AddFile` or `ExecuteCommand`) between `SetGitRepo` and `Commit`?

smuu (changes requested):
This is amazing!

Only one small thing.

e2e/basic/build_from_git_test.go
mojtaba-esk: Good idea, let's do it

smuu (changes requested):
Is it possible to create a check in `SetGitRepo` to check if the selected builder is `kubernetes`?

smuu (approved):

PR #333, opened 2024-05-10
feat: prepare knuu for chaos-mesh

smuu:
## Overview

Unfortunately, I needed to drop the Kubernetes version to `v0.28.2` in order for this to work.

Basically, this PR adds support to extend the `Instance` object to add more use cases without the need for knuu to provision the Kubernetes cluster in some way.


mojtaba-esk (approved):
Awesome

PR #334, opened 2024-05-13
fix: Kaniko builder resource limit

mojtaba-esk:
An issue reported by @rach-id on kaniko build. After investigation, it turned out that it hit the ephemeral storage limit and it seems to be working with `10Gi` when building [Celestia-app](https://github.com/celestiaorg/celestia-app) repo.

This PR suggests a fix to resolve that issue.
More context can be found in [this chat](https://celestia-team.slack.com/archives/C0653SYL6GM/p1715597415962689).

Note: I could not run it on the current version of Robusta as the pod keeps pending for ever to get the resources; tested it on local `minikube`.

### DIFF
--- a/e2e/basic/build_from_git_test.go
+++ b/e2e/basic/build_from_git_test.go
@@ -27,14 +27,14 @@ func TestBuildFromGit(t *testing.T) {
 	instance, err := knuu.NewInstance("my-instance")
 	require.NoError(t, err, "Error creating instance")
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 45*time.Minute)
 	defer cancel()
 
 	// This is a blocking call which builds the image from git repo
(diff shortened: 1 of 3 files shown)


mojtaba-esk:
Merging this PR must wait until we fix the resource provision for it and then it passes the `TestBuildFromGit` test.

smuu (approved):

rach-id (commented):
Ideally, we should have a mechanism that deletes old images if the storage is full. Or even delete all the images. I think all the images would be good for nothing once the tests are over

smuu:
> Ideally, we should have a mechanism that deletes old images if the storage is full. Or even delete all the images. I think all the images would be good for nothing once the tests are over

This is not about images that are not used anymore. We use `ephemeral` storage, which only exists while the Pod runs. As `Kaniko` uses more than we thought, our Kubernetes nodes were unprepared for that. We increased that limit, which should fix it.

By the way, Kubernetes has a garbage collector that deletes unused images if the disk runs full.

PR #336, opened 2024-05-15
feat: reverse proxy

### LINKED ISSUE #315: feat: reverse proxy
- bittwister
- application endpoints (like RPC)
- upcoming GetFile and GetFolder feature

mojtaba-esk:
This is a draft version as there is an issue which needs help from other team members.

To run it, use the following command
```
go test -v ./e2e/basic/ --run TestReverseProxy -timeout 60m
```

Closes #315


smuu:
Could you please provide the commands of how you testet it locally?

smuu:
I started reviewing and to understand it better I have two questions:
1. Does this solution require a reverse proxy (a LoadBalancer) per knuu instance we want to access?
2. Does this solution need to have something pre-installed in the k8s cluster?

pkg/k8s/k8s_pod.go
```diff
@@ -314,6 +315,24 @@ func (c *Client) getPod(ctx context.Context, name string) (*v1.Pod, error) {
...
+		select {
+		case <-ctx.Done():
+			return ErrWaitingForDeployment.WithParams(name).Wrap(err)
+		case <-time.After(waitRetry):
+			// Retry after some seconds
+		}
+	}
+
+	return nil
+}
```
smuu, line 334: Can we move that to a new file called `k8s_deplotment.go`?

pkg/knuu/knuu.go
```diff
@@ -99,15 +101,28 @@ func InitializeWithScope(scope string) error {
...
+	traefikClient = &traefik.Traefik{
+		K8s: k8sClient,
+	}
+	if err := traefikClient.Deploy(ctx); err != nil {
+		return ErrCannotDeployTraefik.Wrap(err)
+	}
+
+	publicIP, err := traefikClient.Endpoint(ctx)
+	if err != nil {
+		fmt.Printf("Error getting traefik Endpoint: %v", err)
```
smuu, line 113: This should return the error instead of just printing it.

pkg/knuu/knuu.go
```diff
@@ -99,15 +101,28 @@ func InitializeWithScope(scope string) error {
...
 	}
+	fmt.Printf("Traefik publicIP: %v\n", publicIP)
```
smuu, line 115: Can we use `logrus.Infof` here?

pkg/traefik/traefik.go
```diff
@@ -0,0 +1,310 @@
...
+	traefikServiceName = "traefik"
+	Port               = 80
+	PortSecure         = 443
+	deploymentName     = "traefik-deployment"
+	roleName           = "traefik-role"
+	containerName      = "traefik"
+	image              = "traefik:v3.0"
+	appLabel           = "app"
+	appLabelValue      = "traefik"
+	replicas           = 2
```
smuu, line 31: Do we actually need two replicas?

pkg/traefik/traefik.go
```diff
@@ -0,0 +1,310 @@
...
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{appLabel: appLabelValue},
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{appLabel: appLabelValue},
+				},
+				Spec: v1.PodSpec{
+					ServiceAccountName: serviceAccountName,
+					Containers: []v1.Container{
```
smuu, line 108: It would be good to add resources (CPU and memory) for traefik to ensure availability

pkg/traefik/traefik.go
```diff
@@ -0,0 +1,310 @@
...
+	if err != nil {
+		return ErrTraefikRoleCreationFailed.Wrap(err)
+	}
+
+	if err := t.K8s.CreateClusterRoleBinding(ctx, clusterRoleName, nil, clusterRoleName, serviceAccountName); err != nil {
+		return ErrTraefikRoleBindingCreationFailed.Wrap(err)
+	}
+
+	// Create the Traefik deployment using the service account
+	traefikDeployment := &appsv1.Deployment{
```
smuu, line 91: Instead of manually deploying the Deployment like that, would it be possible to use a knuu Instance?

pkg/traefik/traefik.go
```diff
@@ -0,0 +1,310 @@
...
+		Version:  "v1alpha1",
+		Resource: "middlewares",
+	}
+
+	_, err := t.K8s.DynamicClient().Resource(middlewareResource).Namespace(t.K8s.Namespace()).Create(ctx, middleware, metav1.CreateOptions{})
+	if err != nil {
+		return ErrTraefikMiddlewareCreationFailed.Wrap(err)
+	}
+	return nil
+}
```
smuu, line 254: Open thought/question:

Knuu was designed with the idea that the Kubernetes Cluster does not need any specific setup (Like installing operators). So you could spin up a new Cluster and run your tests.
For example, chaos-mesh was requested but required an operator, so I created a separate repository. If the dev wants to use chaos-mesh, he/she can import that 'add-on' to the knuu Instance and control the traffic.

I'm not sure how to handle that reverse proxy topic.

Please share your thoughts on that topic @mojtaba-esk @jrmanes

proxy_sa.sh
smuu, line 1: Do we still need this script?

pkg/traefik/traefik.go
```diff
@@ -0,0 +1,310 @@
...
+	// middleware is required to strip the prefix from the service name
+	if err := t.createMiddleware(ctx, prefix, middlewareName); err != nil {
+		return err
+	}
+
+	return t.createIngressRoute(ctx, serviceName, prefix, middlewareName, portTCP)
+}
+
+// TODO: need to update the k8s pkg to handle service creation in more custom way
+func (t *Traefik) createService(ctx context.Context) error {
```
smuu, line 189: If we deploy trafik as a knuu instance, this should be created automatically when ports are added.
The only thing we need to add is a way to make the Service a LoadBalancer type.

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+	select {
+	case <-ctx.Done():
+		return ErrTimeoutWaitingForServiceReady
+	default:
+		return nil
+	}
+}
+
+func (c *Client) GetServiceEndpoint(ctx context.Context, name string) (string, error) {
+	minioService, err := c.clientset.CoreV1().Services(c.namespace).Get(ctx, name, metav1.GetOptions{})
```
smuu, line 191: pls rename `minio` to something more generic as it is used for `traefik` and `minio`

smuu (changes requested):
This is :exploding_head:
Thank you for all the work!
I added some questions and thoughts.

pkg/k8s/k8s_endpoint.go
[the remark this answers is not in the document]
mojtaba-esk: No we can remove it.

pkg/k8s/k8s_pod.go
mojtaba-esk: sure it will be cleaner

pkg/k8s/k8s_service.go
mojtaba-esk: haha copy pasta issue lol

pkg/knuu/knuu.go
mojtaba-esk: nice catch

pkg/knuu/knuu.go
mojtaba-esk: sure even debug

pkg/traefik/traefik.go
mojtaba-esk: I am not sure, probably not for most tests. However, I think we need to make it scalable somehow so if there is a large scale test it can scale up itself.

pkg/traefik/traefik.go
mojtaba-esk: Good point. Added.

pkg/traefik/traefik.go
mojtaba-esk: Good question. We can go either with the add on approach or since right now it is used by the knuu instance when BitTwister is activated or we might add some more fucntions in future like running a custom code, etc... another option is to keep it as a configurable variable so user can activate it when they need it.

MSevey:
> This is a draft version as there is an issue which needs help from other team members.
> 
> To run it, use the following command
> 
> ```
> go test -v ./e2e/basic/ --run TestReverseProxy -timeout 60m
> ```
> 
> Closes #315

Since the issue doesn't have a lot of context, to help with alignment on async reviews, can we add some more detail to either the issue or the PR description? 

Things I'd like to know are:

- What is the current problem that this solves?
- What is the scope of this PR as it relates to the full solution? Is this everything or are there planned next steps? 
- What are the current known limitations or issues?
- Are there any assumptions that this PR is making? 
- Does this change come with any caveats?

e2e/basic/reverse_proxy_test.go
```diff
@@ -0,0 +1,55 @@
...
+	require.NoError(t, main.BitTwister.WaitForStart(ctx), "Error waiting for BitTwister to start")
+
+	// test if BitTwister running in a sidecar is accessible
+	err = main.SetBandwidthLimit(1000)
+	assert.NoError(t, err, "Error setting bandwidth limit")
+
+	// Check if the BitTwister service is set
+	out, err := main.BitTwister.Client().AllServicesStatus()
+	require.NoError(t, err, "Error getting all services status")
+	assert.NotEmpty(t, out, "No services found")
```
MSevey, line 54: If possible it would be good to check for some explicit state other than not empty as this would not protect us from a bug in the code that returns an invalid non empty response.

e2e/basic/reverse_proxy_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"testing"
+	"time"
+
+	"github.com/celestiaorg/knuu/pkg/knuu"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReverseProxy(t *testing.T) {
```
MSevey, line 15: For general tests like this I think it is very helpful to have a comment explaining what we are checking for in the test and why that is helpful/valid. 

Looking at this from a place of having not enough context of the reverse proxy, it isn't clear how this test is testing it and why this is an effective test.

pkg/k8s/k8s.go
```diff
@@ -22,6 +23,9 @@ const (
 
 	// certPath path in the filesystem to the ca.crt
 	certPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+	// waitRetry is the time to wait between retries for a readiness checking
+	waitRetry = 2 * time.Second
```
MSevey, line 28: If there is a reason 2 seconds was picked we should add it to this comment. This is helpful for debugging in the future if 2 seconds causes issues, we have some context to go off of in terms of determining a better value.

pkg/k8s/k8s_role.go
```diff
@@ -29,3 +30,29 @@ func (c *Client) CreateRole(
...
+
+func (c *Client) CreateClusterRole(
+	ctx context.Context,
+	name string,
+	labels map[string]string,
+	policyRules []rbacv1.PolicyRule,
+) error {
+	_, err := c.clientset.RbacV1().ClusterRoles().Get(ctx, name, metav1.GetOptions{})
+	if err == nil || !errors.IsNotFound(err) {
+		return ErrClusterRoleAlreadyExists.WithParams(name)
```
MSevey, line 42: We should include the error since the cases where it is not nil we want to understand the error. 
```suggestion
		return ErrClusterRoleAlreadyExists.WithParams(name, err)
```

pkg/k8s/k8s_role.go
```diff
@@ -29,3 +30,29 @@ func (c *Client) CreateRole(
 func (c *Client) DeleteRole(ctx context.Context, name string) error {
 	return c.clientset.RbacV1().Roles(c.namespace).Delete(ctx, name, metav1.DeleteOptions{})
 }
+
+func (c *Client) CreateClusterRole(
```
MSevey, line 34: This should have a test

pkg/k8s/k8s_role.go
```diff
@@ -29,3 +30,29 @@ func (c *Client) CreateRole(
...
+			Name:   name,
+			Labels: labels,
+		},
+		Rules: policyRules,
+	}
+	_, err = c.clientset.RbacV1().ClusterRoles().Create(ctx, role, metav1.CreateOptions{})
+	return err
+}
+
+func (c *Client) DeleteClusterRole(ctx context.Context, name string) error {
```
MSevey, line 56: This should have a test

pkg/k8s/k8s_rolebinding.go
```diff
@@ -39,3 +40,41 @@ func (c *Client) CreateRoleBinding(
...
+
+func (c *Client) CreateClusterRoleBinding(
+	ctx context.Context,
+	name string,
+	labels map[string]string,
+	clusterRole, serviceAccount string,
+) error {
+	_, err := c.clientset.RbacV1().ClusterRoleBindings().Get(ctx, name, metav1.GetOptions{})
+	if err == nil || !errors.IsNotFound(err) {
+		return ErrClusterRoleBindingAlreadyExists.WithParams(name)
```
MSevey, line 52: same
```suggestion
		return ErrClusterRoleBindingAlreadyExists.WithParams(name, err)
```

pkg/k8s/k8s_rolebinding.go
```diff
@@ -39,3 +40,41 @@ func (c *Client) CreateRoleBinding(
 func (c *Client) DeleteRoleBinding(ctx context.Context, name string) error {
 	return c.clientset.RbacV1().RoleBindings(c.namespace).Delete(ctx, name, metav1.DeleteOptions{})
 }
+
+func (c *Client) CreateClusterRoleBinding(
```
MSevey, line 44: needs a test

pkg/k8s/k8s_rolebinding.go
```diff
@@ -39,3 +40,41 @@ func (c *Client) CreateRoleBinding(
...
+				Namespace: c.namespace,
+			},
+		},
+	}
+
+	_, err = c.clientset.RbacV1().ClusterRoleBindings().Create(ctx, role, metav1.CreateOptions{})
+	return err
+}
+
+func (c *Client) DeleteClusterRoleBinding(ctx context.Context, name string) error {
```
MSevey, line 78: needs a test

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+func (c *Client) WaitForService(ctx context.Context, name string) error {
+	for {
+		service, err := c.clientset.CoreV1().Services(c.namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return ErrGettingService.WithParams(name).Wrap(err)
+		}
+
+		if service.Spec.Type == v1.ServiceTypeLoadBalancer {
+			if len(service.Status.LoadBalancer.Ingress) == 0 {
+				time.Sleep(waitRetry)
```
MSevey, line 157: remove bare sleep, use time.After channel

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+		}
+
+		// Check if service is reachable
+		endpoint, err := c.GetServiceEndpoint(ctx, name)
+		if err != nil {
+			return ErrGettingServiceEndpoint.WithParams(name).Wrap(err)
+		}
+
+		if err := checkServiceConnectivity(endpoint); err != nil {
+			time.Sleep(waitRetry) // Retry after some seconds if the service is not reachable
```
MSevey, line 175: remove bare sleep, use time.After channel

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+		}
+
+		if service.Spec.Type == v1.ServiceTypeLoadBalancer {
+			if len(service.Status.LoadBalancer.Ingress) == 0 {
+				time.Sleep(waitRetry)
+				continue // Wait until the LoadBalancer IP is available
+			}
+		} else if service.Spec.Type == v1.ServiceTypeNodePort {
+			if service.Spec.Ports[0].NodePort == 0 {
+				return ErrNodePortNotSet
```
MSevey, line 162: Same comment on why returning an error and not continuing.

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+			}
+		} else if len(service.Spec.ExternalIPs) == 0 {
+			return ErrExternalIPsNotSet
```
MSevey, line 165: Same comment on why returning an error and not continuing.

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+		}
+
+		// Check if service is reachable
+		endpoint, err := c.GetServiceEndpoint(ctx, name)
+		if err != nil {
+			return ErrGettingServiceEndpoint.WithParams(name).Wrap(err)
```
MSevey, line 171: Same comment on why returning an error and not continuing.

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+		}
+
+		if err := checkServiceConnectivity(endpoint); err != nil {
+			time.Sleep(waitRetry) // Retry after some seconds if the service is not reachable
+			continue
+		}
+
+		break // Service is reachable, exit the loop
```
MSevey, line 179: To have the select statement here instead of the break, I would inverse the above check. 

```go
if err := checkServiceConnectivity(endpoint); err == nil {
        break
}
```

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
 }
+
+func (c *Client) WaitForService(ctx context.Context, name string) error {
+	for {
+		service, err := c.clientset.CoreV1().Services(c.namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return ErrGettingService.WithParams(name).Wrap(err)
+		}
+
+		if service.Spec.Type == v1.ServiceTypeLoadBalancer {
```
MSevey, line 155: These concurrent if conditions make me think a switch statement would be more useful and also asks the question how to handle the case where it is not one of these 2?

```go
switch service.Spec.Type {
case v1.ServiceTypeLoadBalance:
case v1.ServiceTypeNodePort: 
default: // What to do here??
}
```

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+	}
+}
+
+func (c *Client) GetServiceEndpoint(ctx context.Context, name string) (string, error) {
+	srv, err := c.clientset.CoreV1().Services(c.namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return "", ErrGettingService.WithParams(name).Wrap(err)
+	}
+
+	if srv.Spec.Type == v1.ServiceTypeLoadBalancer {
```
MSevey, line 196: Similar comment about using a switch statement to understand what happens if it is not one of these two?

MSevey: I think that would also make it clearer how the return values are being handles as currently the return statement is for this if condition at the bottom, which is kind of far away from the context of the return values.

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+			}
+		}
+		return fmt.Sprintf("%s:%d", nodeIP, srv.Spec.Ports[0].NodePort), nil
+	}
+
+	return fmt.Sprintf("%s:%d", srv.Spec.ClusterIP, srv.Spec.Ports[0].Port), nil
+}
+
+func checkServiceConnectivity(serviceEndpoint string) error {
+	conn, err := net.DialTimeout("tcp", serviceEndpoint, 2*time.Second)
```
MSevey, line 229: more hardcoded time to constant

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
 	}
 	return svc, nil
 }
+
+func (c *Client) WaitForService(ctx context.Context, name string) error {
```
MSevey, line 148: This should have a unit test if possible. If it did I suspect my comments on the for loop not blocking would have been caught.

pkg/knuu/errors.go
MSevey, line 1: Same comment as the follow up on golangci-lint

pkg/traefik/errors.go
```diff
@@ -0,0 +1,50 @@
...
+	if e.Err != nil {
+		return fmt.Sprintf("%s: %v", msg, e.Err)
+	}
+	return msg
+}
+
+func (e *Error) Wrap(err error) error {
+	e.Err = err
+	return e
+}
```
MSevey, line 25: This isn't wrapping, this is replacing. Shouldn't it be

```suggestion
func (e *Error) Wrap(err error) error {
	e.Err = errors.Join(e.Err, err)
	return e
}
```

pkg/traefik/errors.go
MSevey, line 1: Same comment on follow up for golangci

e2e/basic/reverse_proxy_test.go
mojtaba-esk: we could also use an nginx server instead

pkg/k8s/k8s.go
mojtaba-esk: 2 here is just an arbitrary number to wait between the checks. @jrmanes @smuu  what number do you think is the best to be used here?

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+	for {
```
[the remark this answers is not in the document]
mojtaba-esk: nice catch

pkg/traefik/traefik.go
```diff
@@ -28,8 +29,13 @@ const (
...
 	appLabel           = "app"
 	appLabelValue      = "traefik"
-	replicas           = 2
+	replicas           = 1
 	waitRetry          = 5 * time.Second
+
+	defaultCPURequest    = "1000m"
+	defaultMemoryRequest = "512Mi"
+	maxCPULimit          = "2000m"
+	maxMemoryLimit       = "1024Mi"
```
smuu, line 38: Do we need that much of resources?

pkg/k8s/k8s.go
mojtaba-esk: Thanks Jose

e2e/basic/reverse_proxy_test.go
mojtaba-esk: done

e2e/basic/reverse_proxy_test.go
mojtaba-esk: delicate point, updated

pkg/k8s/errors.go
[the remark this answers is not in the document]
mojtaba-esk: yeah using a linter is a great idea in ci. 
About the comments on Exported funcs, I blieve not every function needs it. It is attempted to have clear names for funcs and their arguments so it will be intuitive for users to use it. Otherwise yes we need comments to clarify stuff or even show an example to the users.

pkg/k8s/k8s_rolebinding.go
mojtaba-esk: nice one

pkg/k8s/k8s_service.go
```diff
@@ -142,3 +144,92 @@ func prepareService(
...
+		service, err := c.clientset.CoreV1().Services(c.namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return ErrGettingService.WithParams(name).Wrap(err)
```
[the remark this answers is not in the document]
mojtaba-esk: nice catch

pkg/k8s/k8s_service.go
mojtaba-esk: that's definitely doable. Let's handle it in a seprate PR.

pkg/k8s/k8s_service.go
mojtaba-esk: Thanks for pointing it out. Ill fix it

pkg/k8s/k8s_service.go
mojtaba-esk: fixed

pkg/k8s/k8s_service.go
mojtaba-esk: ok

pkg/k8s/k8s_service.go
mojtaba-esk: Here if there is an error on getting the service like some internal k8s server error or communication error, we would want to propagate it back to the user. The service must be there and it should be queriable because when user calls this function is when a service is created and want to know when it is ready to be used.

pkg/k8s/k8s_service.go
mojtaba-esk: done

pkg/traefik/errors.go
mojtaba-esk: very neat. thanks man

pkg/traefik/traefik.go
mojtaba-esk: Not sure how much we need. How do you think we should configure it? or even make it configurable by the user?

smuu:
In the prototype for big block tests, I added the following function to `Instance`:

```go
func (i *Instance) AddHost(port int) (err error, host string) {
	prefix := fmt.Sprintf("%s-%d", i.k8sName, port)
	if err := traefikClient.AddHost(context.TODO(), i.k8sName, prefix, port); err != nil {
		return ErrAddingHost.WithParams(i.k8sName).Wrap(err), ""
	}
	host, err = traefikClient.URL(context.TODO(), prefix)
	if err != nil {
		return ErrGettingHost.WithParams(i.k8sName).Wrap(err), ""
	}
	return nil, host
}
```

It would be great to have a function with that functionality. Feel free to rename and modify the function at will.

pkg/traefik/traefik.go
smuu: @jrmanes, what's your opinion here?

smuu (approved):

Makefile
MSevey, line 1: Can be a follow up but imo it would be better to DRY up this commands and just use the `pkg` and `run` variables. 

ie:
```bash
pkgs := $(shell go list ./...)
run := .
count := 1

test: 
        KNUU_TIMEOUT=120m go test -v $(pkgs) -run $(run) -count=$(count) -timeout 120m
.PHONY: test
```

then you can just have the one makefile command and easily target whatever by setting the pkg, run, or count flags. 

Targeting a directory
```
make test pkgs=./e2e/basic
```
Targeting a Test in a directory
```
make test pkgs=./e2e/basic run=TestJustThisTest
```
Run a test in a loop to debug
```
make test pkgs=./e2e/basic run=TestJustThisTest10Times count=10
```

This model has more flexibility and DRY's up the Makefile significantly.

pkg/knuu/errors.go
MSevey, line 1: Follow Up. All Error implementations need to update their Wrap function as it isn't wrapping the error it is overriding the error.

pkg/k8s/errors.go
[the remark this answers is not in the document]
MSevey: We can have the discussion around which linters to enable, but we should make a follow up to track getting the linters on the code base.

e2e/basic/reverse_proxy_test.go
MSevey: What was the resolution here?

MSevey (commented):
I don't think my previous review was fully resolved. 

Please don't resolve a conversations unless the action was fully implemented with no deviation. 

Responding to a comment and then dismissing the review by resolving the discussion makes it harder for re-reviews. As it isn't clear if comments were fully addressed so then reviewers have to go through and open up all the discussions to check. Or the same comments will just be left again.

pkg/knuu/errors.go
MSevey: handled by #364 👍🏻

Makefile
mojtaba-esk: that's a good idea. specially when we add unit tests this make file can grow really fast. This way we can have a cleaner code. Let's address it in a separate PR.

pkg/k8s/errors.go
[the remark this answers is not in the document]
mojtaba-esk: an issue is created for it

e2e/basic/reverse_proxy_test.go
mojtaba-esk: https://github.com/celestiaorg/knuu/pull/336/files#diff-c6b5c20a40f4e36e513597308a06b174baa20c53eea0fe97fd057dadb9821ddcR56-R58

mojtaba-esk: instead of just checking if it is empty or not, now we check for existence of a particular field which can yield only if the request goes through successfully.

smuu (approved):

PR #346, opened 2024-05-17
bugfix: sidecar should use parents service

smuu:

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -40,7 +40,7 @@ func (c *CacheOptions) Default(buildContext string) (*CacheOptions, error) {
 		Dir:     "",
 		// ttl.sh with the hash of build context is used as the cache repo
 		// Kaniko adds a string tag to the image name, so we don't need to add it here
-		Repo: fmt.Sprintf("ttl.sh/%s", ctxHash),
+		Repo: fmt.Sprintf("ttl.sh/%s:24h", ctxHash),
 	}, nil
 }
 
--- a/pkg/knuu/instance.go
+++ b/pkg/knuu/instance.go
@@ -743,7 +743,7 @@ func (i *Instance) GetIP() (string, error) {
 	svc, err := k8sClient.GetService(ctx, i.k8sName)
 	if err != nil || svc == nil {
 		// Service does not exist, so we need to deploy it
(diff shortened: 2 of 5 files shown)


pkg/knuu/instance.go
```diff
@@ -743,7 +743,7 @@ func (i *Instance) GetIP() (string, error) {
 	svc, err := k8sClient.GetService(ctx, i.k8sName)
 	if err != nil || svc == nil {
 		// Service does not exist, so we need to deploy it
-		err := i.deployService(ctx)
+		err := i.deployService(ctx, i.portsTCP, i.portsUDP)
```
mojtaba-esk, line 746: This is a good one to pass the ports right here :heart:

pkg/knuu/instance_helper.go
```diff
@@ -80,11 +80,17 @@ func (i *Instance) Labels() map[string]string {
...
+func (i *Instance) deployService(ctx context.Context, portsTCP, portsUDP []int) error {
+	// a sidecar instance should use the parent instance's service
+	if i.isSidecar {
+		return ErrDeployingServiceForSidecar.WithParams(i.k8sName)
+	}
+
+	serviceName := i.k8sName
 	labels := i.getLabels()
-	selectorMap := i.getLabels()
+	labelSelectors := i.getLabels()
```
mojtaba-esk, line 91: since both are the same, we could use either one var or assign one to another. e.g. 

`labelSelectors := labels`

mojtaba-esk (approved):
Well done :rocket:

pkg/builder/builder.go
```diff
@@ -40,7 +40,7 @@ func (c *CacheOptions) Default(buildContext string) (*CacheOptions, error) {
 		Dir:     "",
 		// ttl.sh with the hash of build context is used as the cache repo
 		// Kaniko adds a string tag to the image name, so we don't need to add it here
-		Repo: fmt.Sprintf("ttl.sh/%s", ctxHash),
+		Repo: fmt.Sprintf("ttl.sh/%s:24h", ctxHash),
```
mojtaba-esk, line 43: Nice catch!

mojtaba-esk (approved):

PR #356, opened 2024-05-21
refactor!: have knuu as an object

### LINKED ISSUE #328: refactor!: have knuu as an object
```go
func Knuu(scope string) (*Knuu, error)

func (k *Knuu) GetScope() (string, error)
func (k *Knuu) CleanUp() (error)
func (k *Knuu) SetImageBuilder(b builder.Builder) (error)
func (k *Knuu) ImageBuilder() (builder.Builder, error)
func (k *Knuu) NewInstance(name string) error
```

If we do that before `v1`, please make changes backward compatible.

mojtaba-esk:
This PR proposes a structural change to the knuu pkg to have it as an object.
This helps to have more idiomatic code and enable us to add unit tests easier. Moreover, since every test can have their own knuu object and handle everything in isolation.

Note 1: in order to keep it backward compatible with the current tests, some wrapper functions with a global var is added, and they are in a number of files with `_old.go` postfix so we can delete the files when the users are ready to use this one.

Note 2: There was an attempt not to change everything, i.e. the refactor tries to focus on introducing an object structure for knuu. So, we need a number of refactor rounds on the knuu pkg and in general on the repo.

Closes #328


pkg/knuu/instance_old.go
```diff
@@ -0,0 +1,342 @@
...
+const (
+	None InstanceState = iota
+	Preparing
+	Committed
+	Started
+	Stopped
+	Destroyed
+)
+
+func NewInstance(name string) (*Instance, error) {
```
smuu, line 39: Could this be `func (k *Knuu) NewInstance(name string) (*Instance, error)`?

Then, we could keep a list of all instances created in that scope.

If we have this, we can have something like `knuu.Destroy()` that deletes all instances in that scope.

pkg/knuu/knuu_old.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/celestiaorg/knuu/pkg/builder"
+	"github.com/celestiaorg/knuu/pkg/builder/docker"
+	"github.com/celestiaorg/knuu/pkg/builder/kaniko"
+)
+
+// This is a temporary variable to hold the knuu instance until we refactor knuu pkg
+// TODO: remove this temporary variable
+var tmpKnuu *Knuu
```
MSevey, line 20: Isn't this PR doing what this comment is talking about? 
Why is it still needed?

MSevey:
Since this is a relatively nuanced why of keeping backwards compatibility I would recommend you add some comments to the top of the `_old` files and their corresponding files that captures how the files are setup and should be used while we are in this period of maintaining backwards compatibility. 

Otherwise I think it puts a lot of pressure and burden on the team to keep the context of this refactor in their minds while working on fixes and features.

pkg/knuu/instance_old.go
MSevey: Wouldn't that change need to be added here: https://github.com/celestiaorg/knuu/pull/356/files#diff-3ef3ea3ddb9b6f3d6f910f9da360132513c08de27816ea4956ce5f28d1b08aa2R11

Since this `old` file is for backwards compatibility and shouldn't be changed? 

I would also recommend the convention of `func (k *Knuu) AddInstance()` if the intention of the method is to add an instance to the knuu object.

e2e/basic/build_from_git_test.go
MSevey, line 1: Since this refactor is meant to improve our ability to write unit tests, is there another or a new test that we can create as an example of how this refactor changes our testing?

mojtaba-esk: Well since knuu will be an object and also other dependencies like k8s, minio, ... are also injected as objects into it. It enables us to mock each of them to do a unit test without really deploy something on our clusters. the unit tests will be addressed in separate PRs.

pkg/knuu/instance_old.go
mojtaba-esk: right now we have `knuu.Cleanup()` which we can rename it to `knuu.Destroy()` that deletes the entire namespace and so for that particular case we do not need to keep track of instance objects somewhere in knuu by using `func (k *Knuu) AddInstance()`. 
@smuu @MSevey what other usecases do you think we might want to have by keeping references to instance objects in the knuu object?

smuu:
Is it possible to add a deprecation annotation to the old functions so that we see in the code when we use those functions?

mojtaba-esk:
> Is it possible to add a deprecation annotation to the old functions so that we see in the code when we use those functions?

good idea, done.

pkg/knuu/instance_old.go
MSevey: defer to sam on this one. 
Also we can probably just continue this discussion in a follow up.

e2e/basic/build_from_git_test.go
MSevey: imo this PR should be accompanied by at least 1 unit test to validate the new format. 
since we can't trust the current test suite, as it is broken, we need at least confirmation of one passing test that this refactor achieves the objective.

mojtaba-esk: yes we can add a unit test for it however the main objective was not only unit tests. The redactor's objective mainly is to have more maintainable code and to have more idiomatic code base. and it is not limited only to this refactor we are in the middle a series of refactors.

pkg/knuu/instance_old.go
mojtaba-esk: With the batch processing that we are going to work on, this might come handy.

e2e/basic/build_from_git_test.go
MSevey: I understand that the refactor does other things. 
But there is a general pattern across the org of not prioritizing writing unit tests, that is problematic. 

The expectation and default should be that all PRs have new or updated tests that is maintaining or improving the code coverage of the repository. 

Sufficient code test coverage is a key part of maintainability. If this refactor is aimed at being able to better unit test, and improve maintainability, then both of those things support writing tests.

mojtaba-esk: A unit test is added. @MSevey would you please check it and if everything is ok approve it so we can merge it on Monday?

pkg/knuu/instance_old.go
mojtaba-esk: @smuu any suggestions?

MSevey (approved):
lgtm. I think all open discussions can be turned into follow up issues 👍🏻

MSevey (approved):

Issue #358, opened 2024-05-22
Issue with the volumes


tty47:
related to:
https://github.com/celestiaorg/knuu/actions/runs/9159620991/job/25180413355

```
=== NAME  TestFolderCached
    folder_test_image_cached_test.go:58: Error adding file to 'web10': error copying folder 'resources/html' to instance 'web10: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web7': error copying folder 'resources/html' to instance 'web7: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web5': error copying folder 'resources/html' to instance 'web5: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web2': error copying folder 'resources/html' to instance 'web2: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web8': error copying folder 'resources/html' to instance 'web8: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web3': error copying folder 'resources/html' to instance 'web3: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web9': error copying folder 'resources/html' to instance 'web1: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web1': error copying folder 'resources/html' to instance 'web1: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web6': error copying folder 'resources/html' to instance 'web6: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
    folder_test_image_cached_test.go:58: Error adding file to 'web4': error copying folder 'resources/html' to instance 'web4: the file '/usr/share/nginx/html/index.html' is not a sub folder of any added volume
[...]

PR #359, opened 2024-05-22
fix(volumes): some fixes and cleanups

jrmanes:
hello!

Fixes: https://github.com/celestiaorg/knuu/issues/358

### DIFF
--- a/e2e/basic/file_test_image_cached_test.go
+++ b/e2e/basic/file_test_image_cached_test.go
@@ -37,6 +37,10 @@ func TestFileCached(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Error executing command for '%v': %v", instanceName, err)
 		}
+		err = instance.AddVolumeWithOwner("/usr/share/nginx/html", "1Gi", 0)
+		if err != nil {
+			t.Fatalf("Error adding volume: %v", err)
+		}
 		err = instance.Commit()
 		if err != nil {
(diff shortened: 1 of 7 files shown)


mojtaba-esk (dismissed):
Thanks, well done

smuu (approved):

e2e/basic/file_test_image_cached_test.go
```diff
@@ -37,6 +37,10 @@ func TestFileCached(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Error executing command for '%v': %v", instanceName, err)
 		}
+		err = instance.AddVolumeWithOwner("/usr/share/nginx/html", "1Gi", 0)
```
MSevey, line 40: Anytime we fix something, we should also add a comment. 
Better yet, this entire section of code in this for loop seems like a common test pattern and so should be refactored into a helper. 
Then the helper should be nicely documented explaining why the helper does and why it needs to do it so that future tests can understand how and why to use it.

MSevey (changes requested):
Please refactor into a helper since this was copy pasted into 5 plus tests. 
It is reasonable to believe that future tests will need this same functionality so we should have a helper to DRY up the code.

mojtaba-esk (approved):
:100:

smuu (approved):

MSevey (approved):

Issue #361, opened 2024-05-22
Updathe go doc for SetLatencyAndJitter func


mojtaba-esk:
The current go doc for the following function is misleading users as it is outdated.

func (i *[Instance](https://pkg.go.dev/github.com/celestiaorg/knuu/pkg/knuu@v0.13.2#Instance)) SetLatencyAndJitter(latency, jitter [int64](https://pkg.go.dev/builtin#int64)) [error](https://pkg.go.dev/builtin#error)

The doc indicates that user can set only one of the traffic shaping rules at a time which was a limitation of the very first BitTwister integration. That limitation is no longer the case. so we need to fix it.

more context: https://celestia-team.slack.com/archives/C0653SYL6GM/p1716389298931399

PR #366, opened 2024-05-22
#361 Issue: Update SetLatencyAndJitter func of Go doc

charans29:
## Overview

This pull request updates the documentation for the `SetLatencyAndJitter` function in the `knuu` package. The previous documentation was outdated and misleading, indicating that only one traffic shaping rule could be set at a time and the function could only be called in the 'Committed' state. However, with the current BitTwister integration, these limitations have been removed.

The updates include:
- Clarifying that multiple traffic shaping rules such as latency and jitter can now be set simultaneously.
- Specifying that the function can only be called when the instance is in the 'Started' state.

These changes ensure that the documentation accurately reflects the current capabilities and usage requirements of the `SetLatencyAndJitter` function.

### DIFF
--- a/pkg/knuu/instance.go
+++ b/pkg/knuu/instance.go
@@ -1198,11 +1198,9 @@ func (i *Instance) SetBandwidthLimit(limit int64) error {
 	return nil
 }
 
-// SetLatency sets the latency of the instance
-// latency in ms (e.g. 1000 for 1s)
-// jitter in ms (e.g. 1000 for 1s)
-// Currently, only one of bandwidth, jitter, latency or packet loss can be set
-// This function can only be called in the state 'Commited'
+// SetLatencyAndJitter sets the latency and jitter of an instance in milliseconds (e.g., 1000 for 1 second).
+// Multiple traffic shaping rules like latency, jitter, or packet loss can now be set simultaneously with BitTwister integration.
+// This function can only be called when the instance is in the 'Started' state.
 func (i *Instance) SetLatencyAndJitter(latency, jitter int64) error {
 	if !i.IsInState(Started) {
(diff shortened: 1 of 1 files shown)


pkg/knuu/instance.go
```diff
@@ -1198,11 +1198,9 @@ func (i *Instance) SetBandwidthLimit(limit int64) error {
...
 }
 
-// SetLatency sets the latency of the instance
-// latency in ms (e.g. 1000 for 1s)
-// jitter in ms (e.g. 1000 for 1s)
-// Currently, only one of bandwidth, jitter, latency or packet loss can be set
-// This function can only be called in the state 'Commited'
+// SetLatencyAndJitter sets both the latency and jitter of an instance in milliseconds (e.g., 1000 for 1 second).
+// With the BitTwister integration, multiple traffic shaping rules like latency and jitter can be set simultaneously.
+// This function can only be called when the instance is in the 'Started' state.
```
mojtaba-esk, line 1203: as @staheri14 stated in slack, it will be very useful if we can add some examples here as well. with some explanation what user should expect.

This link provides some documentation about the latency & jitter that is used here: https://www.man7.org/linux/man-pages/man8/tc-netem.8.html

mojtaba-esk (changes requested):

charans29:
I have made the requested changes. Please review them at your convenience. @mojtaba-esk

pkg/knuu/instance.go
charans29: I have made the requested changes. Please review them at your convenience.

mojtaba-esk (approved):

pkg/knuu/instance.go
charans29: You're welcome! It's a pleasure contributing to this project.

PR #373, opened 2024-05-24
Batch processing: Optimize K8S requests

### LINKED ISSUE #357: Optimize K8S requests to don't bulk the K8S API
hello!

currently, we are making a request per resource, so, at scale, this might be an issue, let's try to find a better way to do it
some ideas to avoid doing it in this way:

1. Adjust QPS (Queries Per Second) and Burst Settings
The Kubernetes client-go library allows you to configure the QPS (Queries Per Second) and burst settings for API requests. By default, these settings might be too low for your application’s needs. You can increase them to reduce throttling.
Here's an example of how to adjust these settings:

```go
import (
	"k8s.io/client-go/kubernetes"
	"k8s.io/client-go/rest"
)

func main() {
	config, err := rest.InClusterConfig()
	if err != nil {
		panic(err.Error())
	}

	// Increase QPS and Burst settings
	config.QPS = 100
	config.Burst = 200

	clientset, err := kubernetes.NewForConfig(config)
	if err != nil {
		panic(err.Error())
	}

	// Use clientset to interact with the Kubernetes API
}
```

Adjust the values of QPS and Burst according to your application's requirements and the capacity of your Kubernetes API server.

2. Optimize API Calls
- Review your application logic to see if you can reduce the number of API calls. Some strategies include:
- Batching requests: Combine multiple operations into a single request if possible.
- Caching responses: Cache responses locally to avoid making the same request multiple times.
- Reduce polling frequency: If you are polling the API, consider reducing the frequency of these requests.

3. Aggregating API Requests
While Kubernetes API does not natively support batch operations in a single HTTP request, you can aggregate requests at the application level. This means collecting multiple operations and executing them sequentially or in parallel within the application but with fewer API interactions.
Example: Aggregating Service Creation
Suppose you need to create multiple Kubernetes services. Instead of making individual API calls, you can batch the service creation in your application code.

```go
import (
	"context"
	"fmt"
	"k8s.io/client-go/kubernetes"
	"k8s.io/client-go/rest"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	corev1 "k8s.io/api/core/v1"
)

func main() {
	config, err := rest.InClusterConfig()
	if err != nil {
		panic(err.Error())
	}

	clientset, err := kubernetes.NewForConfig(config)
	if err != nil {
		panic(err.Error())
	}

	services := []corev1.Service{
		{
			ObjectMeta: metav1.ObjectMeta{
				Name: "service1",
			},
			Spec: corev1.ServiceSpec{
				Ports: []corev1.ServicePort{
					{
						Port: 80,
					},
				},
				Selector: map[string]string{
					"app": "app1",
				},
			},
		},
		{
			ObjectMeta: metav1.ObjectMeta{
				Name: "service2",
			},
			Spec: corev1.ServiceSpec{
				Ports: []corev1.ServicePort{
					{
						Port: 80,
					},
				},
				Selector: map[string]string{
					"app": "app2",
				},
			},
		},
	}

	createServices(clientset, "default", services)
}

func createServices(clientset *kubernetes.Clientset, namespace string, services []corev1.Service) {
	for _, service := range services {
		_, err := clientset.CoreV1().Services(namespace).Create(context.TODO(), &service, metav1.CreateOptions{})
		if err != nil {
			fmt.Printf("Error creating service %s: %v\n", service.Name, err)
		} else {
			fmt.Printf("Service %s created successfully\n", service.Name)
		}
	}
}
```

cc: @smuu @mojtaba-esk

mojtaba-esk:
Closes #357

### DIFF
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -22,6 +22,12 @@ const (
 
 	// certPath path in the filesystem to the ca.crt
 	certPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+	// CustomQPS is the QPS to use for the Kubernetes client, DefaultQPS: 5
+	CustomQPS = 100
+
+	// CustomBurst is the Burst to use for the Kubernetes client, DefaultBurst: 10.
+	CustomBurst = 200
 )
 
 type Client struct {
@@ -83,9 +89,21 @@ func isClusterEnvironment() bool {
 }
 
 // getClusterConfig returns the appropriate Kubernetes cluster configuration.
+// If the program is running in a Kubernetes cluster, it returns the in-cluster configuration.
+// Otherwise, it returns the configuration from the kubeconfig file.
+//
+// The QPS and Burst settings are increased to allow for higher throughput and concurrency.
 func getClusterConfig() (*rest.Config, error) {
 	if isClusterEnvironment() {
-		return rest.InClusterConfig()
+		config, err := rest.InClusterConfig()
+		if err != nil {
+			return nil, err
+		}
+
+		// Increase QPS and Burst settings
+		config.QPS = CustomQPS
+		config.Burst = CustomBurst
+		return config, nil
 	}
 
 	// build the configuration from the kubeconfig file
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -8,10 +8,10 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/celestiaorg/knuu/pkg/k8splus"
 	"github.com/sirupsen/logrus"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
@@ -25,10 +25,11 @@ const (
 )
 
 type Client struct {
-	clientset       *kubernetes.Clientset
+	clientset       *k8splus.Clientset
 	discoveryClient *discovery.DiscoveryClient
 	dynamicClient   dynamic.Interface
 	namespace       string
+	useK8sPlus      bool
 }
 
 func New(ctx context.Context, namespace string) (*Client, error) {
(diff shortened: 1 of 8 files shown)


pkg/k8splus/pods.go
```diff
@@ -2,26 +2,52 @@ package k8splus
...
+func (c *pods) CreateBatchPods(ctx context.Context, pods []*v1.Pod, opts metav1.CreateOptions) ([]*v1.Pod, error) {
+	createdPods := make([]*v1.Pod, len(pods))
+	errors := make(chan error, len(pods))
+	var wg sync.WaitGroup
+
+	for i, pod := range pods {
+		wg.Add(1)
+		go func(pod *v1.Pod, index int) {
+			defer wg.Done()
+			createdPod, err := c.PodInterface.Create(ctx, pod, opts)
```
mojtaba-esk, line 37: Nice one; however, this `Create` is calling the API under the hood so we are not batching here. 
We need to create a batch request to the associated k8s endpoint and make it all in one request.
something similar to this one for example:

```go
	podList := &v1.PodList{Items: []v1.Pod{}}
	for _, pod := range pods {
		podList.Items = append(podList.Items, *pod)
	}

	body, err := json.Marshal(podList)
	if err != nil {
		return nil, fmt.Errorf("failed to marshal pods: %w", err)
	}

	req := p.client.Post().
		Namespace(p.ns).
		Resource("pods").
		VersionedParams(&opts, scheme.ParameterCodec).
		Body(body).
		Do(ctx)

	result := &v1.PodList{}
	err = req.Into(result)
	if err != nil {
		return nil, fmt.Errorf("failed to create pods: %w", err)
	}

	createdPods := []*v1.Pod{}
	for _, pod := range result.Items {
		createdPods = append(createdPods, &pod)
	}
```

Not sure if it works, did not test it.

pkg/k8s/k8s.go
```diff
@@ -83,9 +89,21 @@ func isClusterEnvironment() bool {
...
 	if isClusterEnvironment() {
-		return rest.InClusterConfig()
+		config, err := rest.InClusterConfig()
+		if err != nil {
+			return nil, err
+		}
+
+		// Increase QPS and Burst settings
+		config.QPS = CustomQPS
+		config.Burst = CustomBurst
```
mojtaba-esk, line 105: Nice

mojtaba-esk:
We close this PR as, currently, there is no K8s endpoint that receives batch request for creating resources like Replicaset. We explore alternative solutions in separate PRs.

PR #384, opened 2024-05-29
chore: reduce the bittwister tests

### LINKED ISSUE #381: Reduce the number of BitTwister tests
Currently there are a number of sequential tests for BitTwister. The sequential needs to be there as it tests the dynamic traffic shape on one instance. But there are so many and we do not need that for example for Packetlost alone we have the following tests:

```
TestBittwister_Packetloss/10%
TestBittwister_Packetloss/20%
TestBittwister_Packetloss/30%
TestBittwister_Packetloss/50%
TestBittwister_Packetloss/70%
TestBittwister_Packetloss/90%
TestBittwister_Packetloss/100%
```

All of them are not needed, we can probably try 10% and 70% and that's it. so we can have faster e2e tests.

mojtaba-esk:
Closes #381 

## Summary by CodeRabbit

- **Refactor**
  - Updated test cases for bandwidth, packet loss, latency, and jitter to reflect new values and remove outdated scenarios.
  - Refactored test functions to use a test suite structure for improved organization and maintainability.
  - Enhanced instance running state check logic to ensure more reliable instance status verification.

- **Chores**
  - Added setup and teardown logic for test suites to streamline test execution.
  - Updated method signatures to include context parameters for better context handling during cleanup operations.


e2e/basic/bittwister_test.go
```diff
@@ -72,41 +72,17 @@ func TestBittwister_Bandwidth(t *testing.T) {
...
-			targetBandwidth:  1024 * 1000,
-			tolerancePercent: 50,
-		},
+	}{
 		{
-			name:             "2 Mbps",
-			targetBandwidth:  2 * 1024 * 1000,
+			name:             "512 Kbps",
+			targetBandwidth:  512 * 1000,
 			tolerancePercent: 50,
```
MSevey, line 79: 50% deviation is pretty high isn't it? 
if this is a testing limitation then we should have a comment explaining the limitations on testing. 
Otherwise I would think we should have a much lower tolerance, like 10% or something. 

Also you could remove the table test structure now since you just have the one test.

e2e/basic/bittwister_test.go
```diff
@@ -206,31 +182,11 @@ func TestBittwister_Packetloss(t *testing.T) {
 			targetPacketlossRate: 10,
 			tolerancePercent:     50,
 		},
-		{
```
MSevey, line 209: One these ones I would maybe suggest 3 tests. 
1. 0% loss
2. random loss
3. 100% loss

you could then set the random loss to be 1-99.

MSevey: 0 and 100 percent loss to cover edge cases, unless those are not accepted. 
in which case I would do 1% and 99% with the random loss or something.

e2e/basic/bittwister_test.go
MSevey: or you could leave in the table test and have 1 hard coded test and 1 random test.

e2e/basic/bittwister_test.go
```diff
@@ -330,51 +286,11 @@ func TestBittwister_Latency(t *testing.T) {
 			targetLatency:    10 * time.Millisecond,
```
MSevey, line 286: same comments about tolerance and have 1-2 hard coded with a random test.

MSevey:
We should merge this PR in https://github.com/celestiaorg/knuu/pull/401 so that we can run the tests on this PR. 

Is there a reason why this is not against main?

MSevey:
I was looking at the runtimes of the bittwister tests in some recent CI and I think we should restructure these tests even more to save more CI time. 

Currently the test structure is generally the following:
```go
func TestX() {
   t.Parallel()
   // testSetUp
   // define test cases
   // run test cases
}
```

This structure runs the the high level tests in parallel but runs the test cases sequentially. Normally this is ideal because the test setup is the bulk of the test time and the test cases are not very time intensive. 

However for these tests it is the reverse, in that the test cases are where the bulk of the time is spent. 

So we could structure them like this:
```go
func TestX() {
    t.Parallel()
    // Define test cases
    // run testcases
    t.Run(
        t.Parallel()
        // Test setup
    )
}
```

This is going to duplicate the time of test setup up, but enables running each test case in parallel in addition to running each high level test in parallel. 
Running locally the setup for the packetloss test for example is 7s. Adding 7s to the minimum runtime in order to run the test cases in parallel is definitely worth it imo. 

It is worth noting though that this change is optimizing for local CI as the github runners only have a single core I believe so tests never run in parallel anyways. But as it would only increase the overall run time by `n*setupTime` and we are limiting each test to 3 cases based on my previous recommendation, I think it is worth it.

e2e/basic/bittwister_test.go
mojtaba-esk: Well 50% is pretty high, true. Locally we do not need that high tolerance, but when test on cluster, or run a large test, sometimes the tolerance goes higher . Probably it is due to the fact that the time is a factor there.

There are two tests, so it is cleaner to keep the table.

e2e/basic/bittwister_test.go
```diff
@@ -206,31 +182,11 @@ func TestBittwister_Packetloss(t *testing.T) {
 			targetPacketlossRate: 10,
 			tolerancePercent:     50,
```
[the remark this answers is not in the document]
mojtaba-esk: added

e2e/basic/bittwister_test.go
mojtaba-esk: We already had 100% and when we set it, it blocks all packets and the pod is not accessible; therefore, we cannot reach to the pod and measure anything and most of the times k8s API fails to perform anything on that pod.

Setting to `0%` in fact disables the packetloss service for that pod, so not sure if we test anything there.

If we wann go for a random value we should do something between 10% and 90% as the measurement accuracy falls down when we reduce it to near zero or increase it to near 100%.

MSevey (commented):
What are your thoughts on adding the random third test case to help fuzz things? I think this is a value testing strategy.

Also what are your thoughts on the restructure to speed up the tests locally?

mojtaba-esk:
> What are your thoughts on adding the random third test case to help fuzz things? I think this is a value testing strategy.
I think that adds a value, however it is tricky to choose the right tolerance. We either have to keep it a large number for most tests or find a formula to calculate it to let some room for the measurement accuracy errors.

> Also what are your thoughts on the restructure to speed up the tests locally?
I am up for any sort of speedup. What kind of restructuring do you think can speedup things?

smuu (approved):

pkg/instance/instance.go
```diff
@@ -1088,21 +1088,22 @@ func (i *Instance) WaitInstanceIsRunning(ctx context.Context) error {
 	if !i.IsInState(Started) {
 		return ErrWaitingForInstanceNotAllowed.WithParams(i.state.String())
 	}
-	timeout := time.After(1 * time.Minute)
 	tick := time.NewTicker(1 * time.Second)
 
 	for {
+		running, err := i.IsRunning(ctx)
```
mojtaba-esk, line 1094: Since the ticker waits for the defined duration even for the very first trigger, in this example it waits one second before checking if the instance is running or not. Moving the check above the ticker solves it. Although in practice, I am not sure if that helps as usually a pod takes more than a second to be running.

The other change is to use the incoming context instead of having a local timeout.

pkg/knuu/knuu.go
```diff
@@ -169,13 +169,13 @@ func (k *Knuu) CleanUp(ctx context.Context) error {
...
 
-func (k *Knuu) HandleStopSignal() {
+func (k *Knuu) HandleStopSignal(ctx context.Context) {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, syscall.SIGINT, syscall.SIGTERM, os.Interrupt)
 	go func() {
 		<-stop
 		logrus.Info("Received signal to stop, cleaning up resources...")
-		if err := CleanUp(); err != nil {
+		if err := k.CleanUp(ctx); err != nil {
```
mojtaba-esk, line 178: There was a sneaky typo in the code that led to some issues once the test is done.

e2e/bittwister/suite_setup_test.go
```diff
@@ -0,0 +1,41 @@
...
+		ctx = context.Background()
+	)
+	s.Knuu, err = knuu.New(ctx, knuu.WithProxyEnabled())
+	s.Require().NoError(err)
+	s.T().Logf("Scope: %s", s.Knuu.Scope())
+	s.Knuu.HandleStopSignal(ctx)
+}
+
+func (s *Suite) TearDownSuite() {
+	s.T().Cleanup(func() {
```
mojtaba-esk, line 30: Not sure why the Teardown function was called prematurely, so I had to add it to the test cleanup func so it is called once all tests are done.

pkg/instance/instance.go
MSevey: this makes sense, from a code structure standpoint we should always check or perform an action in a loop and then wait. Since in the future things might speed up and sometimes be satisfied on the first loop iteration.

MSevey (approved):
lgtm, I updated the branch to pull in Jose's govuln fix.

smuu (approved):

PR #387, opened 2024-05-29
chore: fix wrapping errors for all pkgs

### LINKED ISSUE #364: fix wrapping errors for all pkgs
This isn't wrapping, this is replacing. Shouldn't it be

```suggestion
func (e *Error) Wrap(err error) error {
	e.Err = errors.Join(e.Err, err)
	return e
}
```

_Originally posted by @MSevey in https://github.com/celestiaorg/knuu/pull/336#discussion_r1608558097_

mojtaba-esk:
Closes #364 

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced error handling with custom error types for more informative and manageable error messages.

- **Tests**
  - Added comprehensive test cases for new error handling features, ensuring robust and reliable error management.


pkg/errors/errors_test.go
```diff
@@ -0,0 +1,133 @@
...
+		},
+		{
+			name: "wrap an error",
+			errorObj: &Error{
+				Message: "test message",
+				Err:     errors.New("initial error"),
+			},
+			wrapError: errors.New("wrapped error"),
+			expected:  "test message: initial error\nwrapped error",
+		},
```
MSevey, line 86: One additional test case could be when the wrapError is a joined error itself. 
Also some opportunity for refactoring common errors
```go
initialError := errors.New("initial error")
wrapError1 := errors.New("wrapped error 1")
wrapError2 := errors.New("wrapped error 2")
...
		{
			name: "wrap a joined error",
			errorObj: &Error{
				Message: "test message",
				Err:     initialError,
			},
			wrapError: errors.Join(wrapError1, WrapError2),
			expected:  "test message: initial error\nwrapped error 1\nwrapped error 2",
		},
```

pkg/errors/errors.go
```diff
@@ -0,0 +1,35 @@
...
+	"errors"
+	"fmt"
+)
+
+type Error struct {
+	Code    string
+	Message string
+	Err     error
+	Params  []interface{}
+}
```
MSevey, line 13: Thank you for refactoring 🙏🏻 

Now I think we could de-export these fields, since we should force interaction via methods.

pkg/errors/errors.go
```diff
@@ -0,0 +1,35 @@
+package errors
+
+import (
+	"errors"
+	"fmt"
+)
+
+type Error struct {
```
MSevey, line 8: Now that we have this refactored out we should add some more context and explanation here. 

Some things that stand out based on current usage:

- There seems to be the implicit expectation that all Errors are used with the `Wrap` function in order to return an `error` type. 
- The Params and Message are linked such that the message is defined initially with an assumed param which must then be known and provided when used. 
- Error() is never used. Is it defined to satisfy the errors interface? If so, then why does Wrap return error and not Error?

pkg/errors/errors_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestError_Error(t *testing.T) {
```
MSevey, line 11: this doesn't test the case of `e.Err = e` I don't think. 

Either we need to add the test case or just remove the check in the `Error()` method. I'm slightly partial to just removing it as I don't think it really saves anything.

pkg/errors/errors_test.go
mojtaba-esk: added a test for it anyways

pkg/errors/errors.go
mojtaba-esk: Nice point

MSevey:
Created an issue with some more thoughts for the future of Errors https://github.com/celestiaorg/knuu/issues/400
Linking here for reference only

pkg/errors/errors.go
mojtaba-esk: done

pkg/errors/errors.go
mojtaba-esk: Well using `Wrap` is an option since this `Error` struct implements `errors.Error`, it can be simply casted implicitly to and `error` type.  But good point, to make it unified and more clear we can make the `Wrap` to return an `*Error` type instead. which can be even more handy for the users.

and yes `Error()` is implemented to satisfy the error interface and it is called when an error gets logged or printed out.

mojtaba-esk: done

pkg/errors/errors_test.go
mojtaba-esk: I also don't think we need to test it . that condition is there to avoid infinite recursion as it happened in one of core/app tests, I added a comment for more clarity.
```go 
if e.err == e {
	return e.message
}
```

MSevey: Can you link to the infinite recursion issue if there is one? 

We should at least update to `errors.Is` for the following lint errror

```
comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error (errorlint)
```

MSevey (changes requested):
last comment about using `errors.Is`

pkg/errors/errors_test.go
mojtaba-esk: let me have a closer look at that lint complain. I was also thinking to add an `Is` method to the errors struct which will be handy for users.

mojtaba-esk: I did not get the same output from my linter with default rules, can you please tell me what rules did you use? so I can reproduce it?

pkg/errors/errors.go
```diff
@@ -19,6 +19,17 @@ func New(code, message string) *Error {
 	}
 }
 
+// It returns true if the error is the same as the given error
+// it checks the error codes for comparison
+func Is(err1, err2 error) bool {
+	if err1 == nil {
+		return false
+	}
```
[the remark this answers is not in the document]
mojtaba-esk: nice catch

pkg/errors/errors_test.go
MSevey: run `make lint` from main with the latest version of golangci-lint. 
The error might not pop up for you, because it is coming from `errCheck` which I ignored for now because there were too many lint errors to address during the initial PR. 

If you enable `errCheck` this should pop up. 

either way you should update to us `errors.Is` because `==` is not a sufficient check for error comparison.

pkg/errors/errors.go
```diff
@@ -0,0 +1,62 @@
...
+// It returns true if the error is the same as the given error
+// it checks the error codes for comparison
+func Is(err1, err2 error) bool {
+	if err1 == nil || err2 == nil {
+		return false
+	}
+	e1, ok1 := err1.(*Error)
+	e2, ok2 := err2.(*Error)
+	return ok1 && ok2 && e1.code == e2.code
+}
```
MSevey, line 31: You shouldn't need to define this, you should use the standard go library `errors.Is` function https://pkg.go.dev/errors#Is

pkg/errors/errors.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (e *Error) Error() string {
+	// We need to keep this condition to avoid infinite recursion
+	if e.err == e {
```
MSevey, line 35:
```suggestion
	if errors.Is(e.err, e) {
```

MSevey (changes requested):

pkg/errors/errors.go
mojtaba-esk: right, that would be more idiomatic

mojtaba-esk: done

MSevey (approved):

PR #401, opened 2024-05-29
fix: testing timeout usage

MSevey:
## Overview

This PR: 

- The merge_group trigger wasn't set so test won't be run on the merge queue. 
- We were setting the KNUU_TIMEOUT specifically in the test, but this wasn't being honored due to the make test command was using the KNUU_TIMEOUT and -timeout variables. 
- ~~I removed the -timeout flags as a default of 10m is sufficient. We have a .env file that we can use locally for the KNUU_TIMEOUT so I removed it from the command itself.~~
- Deleted the .env file 
- I also removed the branch restriction on the PR trigger.

## Summary by CodeRabbit

- **Chores**
  - Increased session timeout from 60 minutes to 120 minutes for improved user experience.
  - Adjusted workflow triggers to enhance CI/CD pipeline efficiency.


.env
```diff
@@ -1,3 +1,3 @@
-KNUU_TIMEOUT=60m # The timeout for the tests
+KNUU_TIMEOUT=120m   # The timeout for the tests
```
mojtaba-esk, line 1: The old version works with this, but the new version of knuu receives it as a parameter so user can pass the duration they need.

.env
```diff
@@ -1,3 +1,3 @@
...
 KNUU_BUILDER=docker # The builder to use for building images. docker or kubernetes
```
mojtaba-esk, line 2: same for this one. and the default now is `kaniko` i.e. k8s based builder

.env
```diff
@@ -1,3 +1,3 @@
...
-LOG_LEVEL=info # The debug level. debug, info, warn, error
+LOG_LEVEL=info      # The debug level. debug, info, warn, error
```
mojtaba-esk, line 3: this one also is a parameter in knuu object and user can set it however they want

Makefile
```diff
@@ -2,6 +2,7 @@ pkgs := $(shell go list ./...)
 run := .
 count := 1
 
+
 test: 
-	KNUU_TIMEOUT=120m go test -v $(pkgs) -run $(run) -count=$(count) -timeout 120m
+	go test -v $(pkgs) -run $(run) -count=$(count)
```
mojtaba-esk, line 7: I would keep that `-timeout 120m` as the default is around `10m` I guess and it is not sufficient to deploy pods and perform a large or medium size test.

mojtaba-esk (changes requested):
It is better to start practice the new knuu version

.env
MSevey: should this just be removed then?

Makefile
MSevey: oh sorry, i was reading that as 120s. my bad will revert.

MSevey: but on the other hand, this is for local knuu tests. 
our test suite should run in under 10m. 

for app, they have their own go test commands where they would need to set their timeout variable accordingly.

.env
MSevey: how does the user pass this param?

.env
mojtaba-esk: I would remove them because that's something to be handled by the user and they should care about their env. The only thing `knuu` does and I am not sure if it is useful anymore is that it loads any `.env` file it finds. @jrmanes knows better about it.

Makefile
mojtaba-esk: yeah ideally. right now even one hour is not enough :)

.env
MSevey: sounds like this is outdate, either way we shouldn't be in the habit of committing .env imo.

mojtaba-esk (dismissed):

.env
smuu: This `.env` file was committed as an example for the developers.

Makefile
```diff
@@ -2,6 +2,7 @@ pkgs := $(shell go list ./...)
 run := .
 count := 1
 
+
 test: 
-	KNUU_TIMEOUT=120m go test -v $(pkgs) -run $(run) -count=$(count) -timeout 120m
```
smuu, line 6: With the `KNUU_TIMEOUT` environment variable, you configure the timeout handler that runs inside the Kubernetes cluster.
With the go test parameter `-timeout`, you define when go stops running the test.

With removing the env variable, knuu will delete everything after the default time (60m).

mojtaba-esk: True and the default time for knuu is also configurable by the user when they wanna create a knuu object.

MSevey: @smuu the reason I removed this here, is that locally 60m should be more than enough and in CI we are setting it again, but if we define it in the `make test` command, the CI value gets ignored. 

imo if we are always defining these env values, then the default should just be that value.

.env
MSevey: It isn't clear that the `.env` was an example as it isn't referenced anywhere. 
Using an `.env` file is pretty common practice so I don't think we need an actual file as an example, just have the env var options clearly documented. 

for future examples, if we need/require a `.env` file because we have things that need to be defined and don't have defaults, we should create a `.env.example`.

mojtaba-esk (approved):

mojtaba-esk (approved):

PR #409, opened 2024-05-30
feat: add support to upload tshark to s3

smuu:

### DIFF
--- a/e2e/tshark/tshark_test.go
+++ b/e2e/tshark/tshark_test.go
@@ -1,0 +1,82 @@
+package basic
+
+import (
+	"context"
+	"log"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+func TestTshark(t *testing.T) {
+	t.Parallel()
+	// Setup
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	knuu, err := knuu.New(ctx)
+	if err != nil {
+		t.Fatalf("Error creating knuu: %v", err)
+	}
+
+	scope := knuu.Scope()
+	log.Println(scope)
+
+	instance, err := knuu.NewInstance("alpine")
+	if err != nil {
+		t.Fatalf("Error creating instance '%v':", err)
+	}
+	err = instance.SetImage(ctx, "docker.io/alpine:latest")
+	if err != nil {
+		t.Fatalf("Error setting image: %v", err)
+	}
+	err = instance.SetCommand("sleep", "infinity")
+	if err != nil {
+		t.Fatalf("Error setting command: %v", err)
+	}
+	err = instance.EnableTsharkCollector(
+		"10Gi", // Example volume size
+		os.Getenv("S3_ACCESS_KEY"),
+		os.Getenv("S3_SECRET_KEY"),
+		os.Getenv("S3_REGION"),
+		os.Getenv("S3_BUCKET_NAME"),
+		"tshark/"+scope,
+	)
+	if err != nil {
+		t.Fatalf("Error enabling tshark collector: %v", err)
+	}
+	err = instance.Commit()
+	if err != nil {
+		t.Fatalf("Error committing instance: %v", err)
+	}
+
+	t.Cleanup(func() {
+		require.NoError(t, instance.Destroy(ctx))
+	})
+
+	// Test logic
+
+	err = instance.Start(ctx)
+	if err != nil {
+		t.Fatalf("Error starting instance: %v", err)
+	}
+	err = instance.WaitInstanceIsRunning(ctx)
+	if err != nil {
+		t.Fatalf("Error waiting for instance to be running: %v", err)
+	}
+	wget, err := instance.ExecuteCommand(ctx, "echo", "Hello World!")
+	if err != nil {
+		t.Fatalf("Error executing command '%v':", err)
+	}
+
+	// wait for 10 minutes
+	time.Sleep(10 * time.Minute)
+
+	assert.Contains(t, wget, "Hello World!")
+}
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -69,6 +69,24 @@ type ObsyConfig struct {
 	prometheusRemoteWriteExporterEndpoint string
 }
 
+// TsharkCollectorConfig represents the configuration for the tshark collector
+type TsharkCollectorConfig struct {
+	// enabled is a flag to enable the tshark collector
+	enabled bool
+	// volumeSize is the size of the volume to use for the tshark collector
+	volumeSize string
+	// s3AccessKey is the access key to use for the s3 server
+	s3AccessKey string
+	// s3SecretKey is the secret key to use for the s3 server
+	s3SecretKey string
+	// s3Region is the region of the s3 server
+	s3Region string
+	// s3Bucket is the bucket to use for the s3 server
+	s3Bucket string
+	// s3KeyPrefix is the key prefix to use for the s3 server
+	s3KeyPrefix string
+}
+
 // SecurityContext represents the security settings for a container
 type SecurityContext struct {
 	// Privileged indicates whether the container should be run in privileged mode
@@ -81,35 +99,36 @@ type SecurityContext struct {
 // Instance represents a instance
 type Instance struct {
 	system.SystemDependencies
-	name                 string
-	imageName            string
-	k8sName              string
-	state                InstanceState
-	instanceType         InstanceType
-	kubernetesService    *v1.Service
-	builderFactory       *container.BuilderFactory
-	kubernetesReplicaSet *appv1.ReplicaSet
-	portsTCP             []int
-	portsUDP             []int
-	command              []string
-	args                 []string
-	env                  map[string]string
-	volumes              []*k8s.Volume
-	memoryRequest        string
-	memoryLimit          string
-	cpuRequest           string
-	policyRules          []rbacv1.PolicyRule
-	livenessProbe        *v1.Probe
-	readinessProbe       *v1.Probe
-	startupProbe         *v1.Probe
-	files                []*k8s.File
-	isSidecar            bool
-	parentInstance       *Instance
-	sidecars             []*Instance
-	fsGroup              int64
-	obsyConfig           *ObsyConfig
-	securityContext      *SecurityContext
-	BitTwister           *btConfig
+	name                  string
+	imageName             string
+	k8sName               string
+	state                 InstanceState
+	instanceType          InstanceType
+	kubernetesService     *v1.Service
+	builderFactory        *container.BuilderFactory
+	kubernetesReplicaSet  *appv1.ReplicaSet
+	portsTCP              []int
+	portsUDP              []int
+	command               []string
+	args                  []string
+	env                   map[string]string
+	volumes               []*k8s.Volume
+	memoryRequest         string
+	memoryLimit           string
+	cpuRequest            string
+	policyRules           []rbacv1.PolicyRule
+	livenessProbe         *v1.Probe
+	readinessProbe        *v1.Probe
+	startupProbe          *v1.Probe
+	files                 []*k8s.File
+	isSidecar             bool
+	parentInstance        *Instance
+	sidecars              []*Instance
+	fsGroup               int64
+	obsyConfig            *ObsyConfig
+	tsharkCollectorConfig *TsharkCollectorConfig
+	securityContext       *SecurityContext
+	BitTwister            *btConfig
 }
 
 func New(name string, sysDeps system.SystemDependencies) (*Instance, error) {
@@ -134,6 +153,14 @@ func New(name string, sysDeps system.SystemDependencies) (*Instance, error) {
 		prometheusExporterEndpoint:            "",
 		prometheusRemoteWriteExporterEndpoint: "",
 	}
+	tsharkCollectorConfig := &TsharkCollectorConfig{
+		enabled:     false,
+		s3AccessKey: "",
+		s3SecretKey: "",
+		s3Region:    "",
+		s3Bucket:    "",
+		s3KeyPrefix: "",
+	}
 	securityContext := &SecurityContext{
 		privileged:      false,
 		capabilitiesAdd: make([]string, 0),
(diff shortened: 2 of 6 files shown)


e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,82 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	knuu, err := knuu.New(ctx)
+	if err != nil {
+		t.Fatalf("Error creating knuu: %v", err)
+	}
+
+	scope := knuu.Scope()
+	log.Println(scope)
```
MSevey, line 29: fwiw, in testing you can use `t.Log()`.

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		t.Fatalf("Error waiting for instance to be running: %v", err)
+	}
+	wget, err := instance.ExecuteCommand(ctx, "echo", "Hello World!")
+	if err != nil {
+		t.Fatalf("Error executing command '%v':", err)
+	}
+
+	// wait for 10 minutes
+	time.Sleep(10 * time.Minute)
```
MSevey, line 79: why are we waiting for 10 minutes in testing? 
We should never have bare sleeps in testing as that just leads to unnecessarily long CI. 

Instead we should be actively probing for something.

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	assert.Contains(t, wget, "Hello World!")
+}
```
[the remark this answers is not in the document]
MSevey: yes, this.

pkg/instance/instance.go
```diff
@@ -134,39 +153,48 @@ func New(name string, sysDeps system.SystemDependencies) (*Instance, error) {
 		prometheusExporterEndpoint:            "",
 		prometheusRemoteWriteExporterEndpoint: "",
 	}
+	tsharkCollectorConfig := &TsharkCollectorConfig{
```
MSevey, line 156: yea on New, we could just set this to nil.

pkg/instance/instance.go
```diff
@@ -959,6 +987,24 @@ func (i *Instance) SetPrometheusRemoteWriteExporter(endpoint string) error {
 	return nil
 }
 
+func (i *Instance) EnableTsharkCollector(volumeSize, s3AccessKey, s3SecretKey, s3Region, s3BucketName, s3KeyPrefix string) error {
```
MSevey, line 990: then enable can be updated to check if not nil to prevent multiple calls overwriting things by accident.

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,82 @@
...
+	t.Parallel()
+	// Setup
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	knuu, err := knuu.New(ctx)
+	if err != nil {
+		t.Fatalf("Error creating knuu: %v", err)
+	}
```
mojtaba-esk, line 26: instead of this I suggest to use `require.NoError(t, err, "optional text")` it is more concise.

pkg/instance/tshark.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+)
+
+func (i *Instance) createTsharkCollectorInstance(ctx context.Context) (*Instance, error) {
+	tsharkCollector, err := New("tshark-collector", i.SystemDependencies)
+	if err != nil {
+		return nil, err
+	}
+	if tsharkCollector.SetImage(ctx, "ttl.sh/68d00fc2-c488-4e3a-a787-5c0e8c8135d9:24h") != nil {
```
mojtaba-esk, line 12: I think there is a typo here. this image lasts for 24 hours and then will be deleted.

pkg/instance/tshark.go
```diff
@@ -0,0 +1,49 @@
+package instance
+
+import (
+	"context"
+)
+
+func (i *Instance) createTsharkCollectorInstance(ctx context.Context) (*Instance, error) {
+	tsharkCollector, err := New("tshark-collector", i.SystemDependencies)
```
mojtaba-esk, line 8: Consider extracting constants

pkg/instance/tshark.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	if tsharkCollector.SetEnvironmentVariable("STORAGE_BUCKET_NAME", i.tsharkCollectorConfig.s3Bucket) != nil {
+		return nil, err
+	}
+	if tsharkCollector.SetEnvironmentVariable("STORAGE_KEY_PREFIX", i.tsharkCollectorConfig.s3KeyPrefix) != nil {
+		return nil, err
+	}
+	if tsharkCollector.SetEnvironmentVariable("CAPTURE_FILE_NAME", i.k8sName+".pcapng") != nil {
+		return nil, err
+	}
```
mojtaba-esk, line 44: These could also be written in a loop

pkg/knuu/knuu_old.go
```diff
@@ -96,7 +96,7 @@ func InitializeWithScope(testScope string) error {
 	}
 
 	// TODO: this must be moved to somewhere more meaningful
-	tmpKnuu.HandleStopSignal()
+	// tmpKnuu.HandleStopSignal()
```
mojtaba-esk, line 99: why this line is commented out? does it create any problems for this feature?

mojtaba-esk (changes requested):
well done, looks good.

e2e/tshark/tshark_test.go
MSevey: discussed offline, this is leftover from debugging, should be removed.

MSevey (changes requested):
only required change is removal of the 10 min sleep in the test. 
everything else can be discussed more in followups.

pkg/knuu/knuu_old.go
smuu: For my workflow, it's annoying to wait for the deletion to complete. I will enable it again.

Instead of cleaning up like that, could we deploy a timeout handler without a time specified so that the cleanup is async?

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,69 @@
...
+	// Setup
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	knuu, err := knuu.New(ctx)
+	require.NoError(t, err, "Error creating knuu")
+
+	scope := knuu.Scope()
+	t.Log(scope)
```
mojtaba-esk, line 26: [nit] No need to log it here as it is already logged in the `knuu.New`

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,69 @@
...
+	err = instance.SetCommand("sleep", "infinity")
+	require.NoError(t, err, "Error setting command")
+
+	err = instance.EnableTsharkCollector(
+		"10Gi", // Example volume size
+		os.Getenv("S3_ACCESS_KEY"),
+		os.Getenv("S3_SECRET_KEY"),
+		os.Getenv("S3_REGION"),
+		os.Getenv("S3_BUCKET_NAME"),
+		"tshark/"+scope,
```
mojtaba-esk, line 43: would be nice to extract constants

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,69 @@
...
+	defer cancel()
+
+	knuu, err := knuu.New(ctx)
+	require.NoError(t, err, "Error creating knuu")
+
+	scope := knuu.Scope()
+	t.Log(scope)
+
+	instance, err := knuu.NewInstance("alpine")
+	require.NoError(t, err, "Error creating instance")
```
mojtaba-esk, line 29: Nice!

pkg/instance/instance.go
```diff
@@ -69,6 +69,22 @@ type ObsyConfig struct {
 	prometheusRemoteWriteExporterEndpoint string
 }
 
+// TsharkCollectorConfig represents the configuration for the tshark collector
+type TsharkCollectorConfig struct {
+	// volumeSize is the size of the volume to use for the tshark collector
+	volumeSize string
```
mojtaba-esk, line 75: Is it possible to change this type to resource.Quantity ? like the one that comes out of  this func: https://pkg.go.dev/k8s.io/apimachinery@v0.28.2/pkg/api/resource#ParseQuantity

then it will be less prone to errors

pkg/instance/instance.go
```diff
@@ -959,6 +977,37 @@ func (i *Instance) SetPrometheusRemoteWriteExporter(endpoint string) error {
...
 }
 
+// TsharkCollectorEnabled returns true if the tshark collector is enabled
+func (i *Instance) TsharkCollectorEnabled() bool {
+	return i.tsharkCollectorConfig != nil
+}
+
+// EnableTsharkCollector enables the tshark collector for the instance
+// This function can only be called in the state 'Preparing' or 'Committed'
+func (i *Instance) EnableTsharkCollector(volumeSize, s3AccessKey, s3SecretKey, s3Region, s3BucketName, s3KeyPrefix string) error {
```
mojtaba-esk, line 987: Would be better to define a config struct for this and pass it instead of too many arguments. That way we can also change the type, order and/or the number of arguments without braking the code in future,

pkg/instance/instance.go
```diff
@@ -959,6 +977,37 @@ func (i *Instance) SetPrometheusRemoteWriteExporter(endpoint string) error {
...
+		return ErrTsharkCollectorConfigNotSet.WithParams(volumeSize, "REDACTED", "REDACTED", s3Region, s3BucketName)
+	}
+	i.tsharkCollectorConfig = &TsharkCollectorConfig{
+		volumeSize:  volumeSize,
+		s3AccessKey: s3AccessKey,
+		s3SecretKey: s3SecretKey,
+		s3Region:    s3Region,
+		s3Bucket:    s3BucketName,
+		s3KeyPrefix: s3KeyPrefix,
+	}
```
mojtaba-esk, line 1006: Exactly this is a good one to pass this struct directly to this func instead.

pkg/instance/tshark.go
```diff
@@ -0,0 +1,54 @@
...
+	if tsharkCollector.AddVolume(tsharkCollectorVolumePath, i.tsharkCollectorConfig.volumeSize) != nil {
+		return nil, err
+	}
+	envVars := map[string]string{
+		"STORAGE_ACCESS_KEY_ID":     i.tsharkCollectorConfig.s3AccessKey,
+		"STORAGE_SECRET_ACCESS_KEY": i.tsharkCollectorConfig.s3SecretKey,
+		"STORAGE_REGION":            i.tsharkCollectorConfig.s3Region,
+		"STORAGE_BUCKET_NAME":       i.tsharkCollectorConfig.s3Bucket,
+		"STORAGE_KEY_PREFIX":        i.tsharkCollectorConfig.s3KeyPrefix,
+		"CAPTURE_FILE_NAME":         i.k8sName + ".pcapng",
```
mojtaba-esk, line 42: we could also extract these key names

pkg/instance/tshark.go
```diff
@@ -0,0 +1,54 @@
...
+	if tsharkCollector.Commit() != nil {
+		return nil, err
+	}
+	if tsharkCollector.SetCPU(tsharkCollectorCPU) != nil {
+		return nil, err
+	}
+	if tsharkCollector.SetMemory(tsharkCollectorMemory, tsharkCollectorMemory) != nil {
+		return nil, err
+	}
+	if tsharkCollector.AddVolume(tsharkCollectorVolumePath, i.tsharkCollectorConfig.volumeSize) != nil {
```
mojtaba-esk, line 33: Would also be nice to refactor these function to receive resource.Quantity type instead. We can do it in a follow up PR.

mojtaba-esk (changes requested):
Well done !
left some suggestions.

pkg/knuu/knuu_old.go
mojtaba-esk: I think to have a better structured tests, we can use test suit and add the cleanup in the tear down func. Also this handle signal can be added by the user as well. So currently the new knuu obj does not call it by default.

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,69 @@
...
+	err = instance.WaitInstanceIsRunning(ctx)
+	require.NoError(t, err, "Error waiting for instance to be running")
+
+	wget, err := instance.ExecuteCommand(ctx, "echo", "Hello World!")
+	require.NoError(t, err, "Error executing command")
+
+	// wait for 1 minute to upload network traces to s3
+	time.Sleep(1 * time.Minute)
+
+	assert.Contains(t, wget, "Hello World!")
```
MSevey, line 68: It isn't clear why you need this sleep? 
If this is purely for debugging purposes, I'd comment it out and update the comment to specifically explain using it for debugging. 
The assertion on `wget` should be deterministic from line 62 and not impacted by a sleep after the command is executed.

pkg/instance/instance.go
MSevey: good follow up imo

pkg/instance/instance.go
```diff
@@ -959,6 +977,37 @@ func (i *Instance) SetPrometheusRemoteWriteExporter(endpoint string) error {
...
+	// volumeSize, s3AccessKey, s3SecretKey, s3Region, s3BucketName, s3KeyPrefix string
+	if err := i.validateStateForObsy("Tshark collector"); err != nil {
+		return err
+	}
+	if i.TsharkCollectorEnabled() {
+		return ErrTsharkCollectorAlreadyEnabled
+	}
+	// prefix can be empty, so it's not needed to check it
+	if volumeSize == "" || s3AccessKey == "" || s3SecretKey == "" || s3Region == "" || s3BucketName == "" {
+		return ErrTsharkCollectorConfigNotSet.WithParams(volumeSize, "REDACTED", "REDACTED", s3Region, s3BucketName)
```
MSevey, line 997: [non-blocking] if we never return the access and secret key we should just removed them from the error message. Then when we are debugging or testing we can just add a log statement. That would prevent accidental logging of these keys in the future if we forgot to revert the `REDACTED` by accident.

MSevey (dismissed):
lgtm, some follow ups though that don't need to block merging this in.

e2e/tshark/tshark_test.go
smuu: This test checks if the `tshark` data is successfully uploaded to s3. This I have done manually.
So in theory we should check if a file was created at the expected path and is updated regularly.

pkg/instance/tshark.go
mojtaba-esk: issue created #426

e2e/tshark/tshark_test.go
MSevey: If there is a way to programmatically check for the file, we should do that in the test. 

If there isn't (as of today) we should turn this into a follow up and figure out how to do it. I'd also add a TODO comment here in the test that we need to check the file so that is clear.

Either way the sleep should be removed.

pkg/instance/instance.go
```diff
@@ -81,35 +102,36 @@
...
+	livenessProbe         *v1.Probe
+	readinessProbe        *v1.Probe
+	startupProbe          *v1.Probe
+	files                 []*k8s.File
+	isSidecar             bool
+	parentInstance        *Instance
+	sidecars              []*Instance
+	fsGroup               int64
+	obsyConfig            *ObsyConfig
+	tsharkCollectorConfig *TsharkCollectorConfig
```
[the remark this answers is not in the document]
mojtaba-esk: thanks for pointing it out.  I fixed it in a new commit.

pkg/instance/tshark.go
```diff
@@ -46,6 +46,10 @@ func (i *Instance) createTsharkCollectorInstance(ctx context.Context) (*Instance
 		return nil, err
 	}
 
+	if i.tsharkCollectorConfig == nil {
+		return nil, ErrTsharkCollectorConfigNotSet
+	}
+
```
MSevey, line 52: This needs to be before `AddVolume` as well since it uses `i.tsharkCollectorConfig.VolumeSize`. 
imo this should be the first check of the method to prevent unnecessary actions prior to such a baseline requirement.

MSevey (commented):
I think most things are resolved? Still a lot of unresolved conversations that are listed as outdated now with the new changes, but some aren't addressed. 

Please fix and resolve, or create an issue and resolve open discussions. 

@mojtaba-esk can you resolve the ones you fixed with your recent push? I resolved some that I believe were fixed. 

I think the coderabbitai comment on validating the config is valid.

MSevey (dismissed):

e2e/tshark/tshark_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/celestiaorg/knuu/pkg/instance"
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+const (
+	s3BucketName = "tshark-test-bucket"
+	s3Location   = "eu-east-1"
+)
```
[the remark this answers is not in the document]
mojtaba-esk: that would be a good idea

mojtaba-esk (approved):

MSevey (dismissed):

smuu:
Thanks @mojtaba-esk for driving this home!

MSevey (approved):

PR #416, opened 2024-06-04
chore: unittest for k8s pkg

### LINKED ISSUE #360: Unit test for k8s pkg functions
This should have a unit test if possible. If it did I suspect my comments on the for loop not blocking would have been caught.

_Originally posted by @MSevey in https://github.com/celestiaorg/knuu/pull/336#discussion_r1608553266_

mojtaba-esk:
Closes #360 

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive test functions for Kubernetes ConfigMaps, DaemonSets, Namespaces, and Network Policies.
  - Introduced a test suite for setting up a fake Kubernetes client and testing various Kubernetes resources.
  - Added utility functions for checking cluster environments, retrieving configurations, and sanitizing resource names, along with their respective tests.


pkg/k8s/k8s.go
```diff
@@ -81,7 +75,21 @@ func New(ctx context.Context, namespace string) (*Client, error) {
 	return kc, nil
 }
 
-func (c *Client) Clientset() *kubernetes.Clientset {
+func NewCustom(
```
MSevey, line 78: Few comments on naming conventions. 

`NewCustom` implies there is a `New` function that could be called that returns some default. 
Since there aren't defaults, and what is currently being supplied are the minimum requirements, I would call this `NewClient`.  

We should also always include the name of what is being returned to be explicit. Especially since our `k8` package has a bunch of structs defined that could have `New` methods.

pkg/k8s/k8s_service.go
```diff
@@ -149,33 +149,36 @@ func (c *Client) WaitForService(ctx context.Context, name string) error {
...
-				continue
+	for firstIteration := true; ; firstIteration = false {
+		if !firstIteration {
+			select {
+			case <-ctx.Done():
+				return ErrTimeoutWaitingForServiceReady
+			case <-ticker.C:
+				// Wait for the next tick before proceeding
 			}
+		}
```
MSevey, line 160: if it isn't important to use a ticker, i.e. checking on a strict interval, then we could just use `time.After` and avoid using this `firstIteration` variable. 
The additional benefit of the `time.After` is that we could have custom intervals based on the retry reason. 

```suggestion
        retryInterval := 0
	for  {
			select {
			case <-ctx.Done():
				return ErrTimeoutWaitingForServiceReady
			case <-time.After(retryInterval):
			        // Reset to default interval
			        retryInterval = waitRetry
			}
```

pkg/k8s/utils.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// build the configuration from the kubeconfig file
+	kubeconfig := filepath.Join(os.Getenv("HOME"), ".kube", "config")
+	return clientcmd.BuildConfigFromFlags("", kubeconfig)
+}
+
+// precompile the regular expression to avoid recompiling it on every function call
+var invalidCharsRegexp = regexp.MustCompile(`[^a-z0-9-]+`)
+
+// SanitizeName ensures compliance with Kubernetes DNS-1123 subdomain names. It:
```
MSevey, line 37: is an empty string a valid name? if not we might want to panic on `len(sanitized) == 0`

MSevey (changes requested):
Great work!  Left a few comments I think are worth cleaning up. 

Also the bittwister tests are failing. I recommend capturing the failure in an issue to see if it is just a one off flaky test or if that random combination is deterministic. 

Lastly, I think it would be helpful to pull all the small helpers you created in the test files into the `suite_setup_test.go` file. Maybe just call it `test_suite_test.go` or something, so that it is easy for people to find the common helpers for future tests.

pkg/builder/kaniko/kaniko.go
```diff
@@ -31,10 +31,9 @@ const (
 )
 
 type Kaniko struct {
-	K8sClientset kubernetes.Interface
-	K8sNamespace string
-	Minio        *minio.Minio // Minio service to store the build context if it's a directory
-	ContentName  string       // Name of the content pushed to Minio
+	K8s         k8s.KubeManager
```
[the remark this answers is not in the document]
MSevey: well `K8S` vs `k8s` depends on if it needs to be exported or not.

MSevey:
Also I don't think these tests run. I just tried searching the CI logs for one of the tests and it doesn't show up.

pkg/builder/kaniko/kaniko.go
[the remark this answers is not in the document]
mojtaba-esk: good point. since this struct is initialized by the user, it has to be exported

pkg/builder/kaniko/kaniko_test.go
```diff
@@ -14,20 +14,21 @@ import (
...
 const (
-	k8sNamespace = "test-namespace"
+	k8sNamespace    = "test-namespace"
+	testImage       = "test-image"
+	testDestination = "registry.example.com/test-image:latest"
 )
 
 func TestKanikoBuilder(t *testing.T) {
 	k8sCS := fake.NewSimpleClientset()
 	kb := &Kaniko{
```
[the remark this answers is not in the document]
mojtaba-esk: it is defined in the same package

pkg/k8s/k8s.go
mojtaba-esk: good point.

mojtaba-esk: well I think the return type is disruptive enough:

```go
func New(ctx context.Context, namespace string) (*Client, error)
```
there is only one struct called `Client` .
would you mind to elaborate more on this point?

mojtaba-esk: and that `New` fucntion initilizes the required values by itself. and that `NewCustom` does not. which is used mostly for tests:
```go
func NewCustom(
	cs kubernetes.Interface,
	dc discovery.DiscoveryInterface,
	dC dynamic.Interface,
	namespace string,
) *Client 
```

so how do you think we can make this more clear?

pkg/k8s/k8s_service.go
mojtaba-esk: we can also use `time.After` but still without `firstIteration` in this particular case, it waits for the `retryInterval` and then checks is the service is ready. The point of having that `firstIteration` is to check it first and then start the waiting.

pkg/k8s/types.go
```diff
@@ -9,12 +9,13 @@ import (
...
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 )
 
 type KubeManager interface {
-	Clientset() *kubernetes.Clientset
+	Clientset() kubernetes.Interface
```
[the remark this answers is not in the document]
mojtaba-esk: The previous one was not an interface and this one allows us to mock it for testing purpose.

pkg/k8s/utils.go
```diff
@@ -0,0 +1,57 @@
...
+	return fileExists(tokenPath) && fileExists(certPath)
+}
+
+func fileExists(path string) bool {
+	_, err := os.Stat(path)
+	return err == nil
+}
+
+// getClusterConfig returns the appropriate Kubernetes cluster configuration.
+func getClusterConfig() (*rest.Config, error) {
```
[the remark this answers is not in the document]
mojtaba-esk: yes it is and moved to `utils.go` file for a better organization.

pkg/k8s/utils.go
```diff
@@ -0,0 +1,57 @@
...
+	if isClusterEnvironment() {
+		return rest.InClusterConfig()
+	}
+
+	// build the configuration from the kubeconfig file
+	kubeconfig := filepath.Join(os.Getenv("HOME"), ".kube", "config")
```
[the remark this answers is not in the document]
mojtaba-esk: `getClusterConfig()` is called only once and it is when the k8s client is initialized with the `New` func. So not sure if defining as a global var helps it.

pkg/k8s/utils.go
mojtaba-esk: Nice catch. will add it. However, not sure if we should panic. we might want to add a random string instead of an empty one. Not really a big fan of panic for this simple thing.
@jrmanes @smuu what do you think?

mojtaba-esk: we could also return a string like this: `empty-name`

mojtaba-esk: well in theory yes, we might. In practice, the output of this function goes to another func that generates a name with a random suffix. If we call that one, then we need to return an error here.

pkg/k8s/k8s.go
MSevey: We shouldn't rely on return types as the descriptor because that isn't what people are reading first.  Developers are going to be typing out `k8.New...`.  When naming we should think about the developer using the method, not us, the developers building it.  1k8.New` and `k8.NewCustom` aren't clear enough imo. 

The k8 package has lots of different types which is why we need the descriptor in the name, otherwise no other type could have a `New` method. 

This is how `New` and `NewCustom` methods should interact
```go 
func NewClient(ctx context.Context, namespace string) (*Client, error) {
	config, err := getClusterConfig()
	if err != nil {
		return nil, ErrRetrievingKubernetesConfig.Wrap(err)
	}

	cs, err := kubernetes.NewForConfig(config)
	if err != nil {
		return nil, ErrCreatingClientset.Wrap(err)
	}

	// create discovery client
	dc, err := discovery.NewDiscoveryClientForConfig(config)
	if err != nil {
		return nil, ErrCreatingDiscoveryClient.Wrap(err)
	}

	// Create the dynamic client
	dC, err := dynamic.NewForConfig(config)
	if err != nil {
		return nil, ErrCreatingDynamicClient.Wrap(err)
	}
	return NewClientCustom(cs, dc, dC, namespace)
}

func NewClientCustom(
	cs kubernetes.Interface,
	dc discovery.DiscoveryInterface,
	dC dynamic.Interface,
	namespace string,
) (*Client, error) {
	kc := &Client{clientset: cs, discoveryClient: dc, dynamicClient: dC}

	namespace = SanitizeName(namespace)
	kc.namespace = namespace
	if kc.NamespaceExists(ctx, namespace) {
		logrus.Debugf("Namespace %s already exists, continuing.\n", namespace)
		return kc, nil
	}

	if err := kc.CreateNamespace(ctx, namespace); err != nil {
		return nil, ErrCreatingNamespace.WithParams(namespace).Wrap(err)
	}

	return kc, nil
}
```

pkg/k8s/utils.go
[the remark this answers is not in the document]
MSevey: It is called in `PortForwardPod` and `RunCommandInPod` as well

pkg/k8s/k8s_service.go
MSevey: this is why my example starts with `retryInterval = 0`

MSevey:
@coderabbitai resolve

MSevey:
@coderabbitai ignore

MSevey:
@coderabbitai pause

MSevey (commented):
my previous review comments should be addressed

pkg/k8s/k8s_service.go
mojtaba-esk: got it, nice trick

pkg/k8s/utils.go
mojtaba-esk: yes that makes sense

MSevey (dismissed):

MSevey (approved):

PR #419, opened 2024-06-04
chore: refactor minio and kaniko to new k8s obj

mojtaba-esk:
This PR proposes a refactor to update Minio and Kaniko pkgs to use K8s new object instead of having clinetset & namespace separately.

### DIFF
--- a/pkg/minio/minio.go
+++ b/pkg/minio/minio.go
@@ -4,8 +4,9 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"net"
 	"time"
+
+	"github.com/celestiaorg/knuu/pkg/k8s"
 
 	miniogo "github.com/minio/minio-go/v7"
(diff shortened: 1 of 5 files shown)


pkg/minio/minio.go
```diff
@@ -368,64 +367,13 @@ func (m *Minio) waitForMinio(ctx context.Context) error {
...
-}
-
-func checkServiceConnectivity(serviceEndpoint string) error {
-	conn, err := net.DialTimeout("tcp", serviceEndpoint, 2*time.Second)
-	if err != nil {
-		return ErrMinioFailedToConnect.WithParams(serviceEndpoint).Wrap(err)
-	}
-	defer conn.Close()
-	return nil // success
-}
```
smuu, line 420: Why did you remove these?

mojtaba-esk: because now we have a func in k8s package that provides `waitForService` and does the same thing.

Issue #377, opened 2024-05-28
Move logger to a `log` folder


tty47:
based on: https://github.com/celestiaorg/knuu/pull/356#discussion_r1617380420

move the `defaultLogger` function to a log directory and use it from the other packages

mahiro72:
Hello @jrmanes , I'm interested in working on this issue. Would it be okay for me to pick it up?

PR #421, opened 2024-06-05
fix: move defaultLogger to a log directory

mahiro72:
## Overview
- move the `defaultLogger` function to a log directory and rename `DefaultLogger`
- use `DefaultLogger` from the other packages

issue:
- https://github.com/celestiaorg/knuu/issues/377


mojtaba-esk (approved):
Great job!

PR #434, opened 2024-06-10
chore!: make minio client optional and improve initialization

### LINKED ISSUE #429: Optimize Minio Client Usage by Reusing Across Requests
## Description

Currently, our application creates a new Minio client (`miniogo.Client`) for each request. This approach can lead to performance inefficiencies due to the overhead associated with repeatedly initializing new clients, especially under high load.

## Proposed Change

Refactor the application to instantiate a single Minio client and reuse it across multiple requests. This change aims to reduce initialization overhead and improve overall performance by leveraging connection pooling and reducing setup times.

## Benefits

- **Reduced Latency:** Minimizes the time spent in client setup for each request.
- **Improved Resource Utilization:** Lessens the burden on system resources by reducing the number of client instances.
- **Enhanced Performance:** Potentially increases throughput by reusing existing HTTP connections.

## Implementation Outline

1. **Create a `MinioService` struct** that includes a `miniogo.Client` as a field.
2. **Instantiate `MinioService`** at application startup with a single Minio client.
3. **Refactor existing code** to use the Minio client from `MinioService` instead of creating new instances.

## Code Snippet

```go
type MinioService struct {
    Client miniogo.Client
}
func NewMinioService(endpoint, accessKey, secretKey string) (MinioService, error) {
    cli, err := miniogo.New(endpoint, &miniogo.Options{
        Creds: credentials.NewStaticV4(accessKey, secretKey, ""),
        Secure: false,
})
if err != nil {
   return nil, err
}
return &MinioService{
       Client: cli,
    }, nil
}
```

mojtaba-esk:
Closes #429 

## Summary by CodeRabbit

- **New Features**
  - Introduced a consistent Minio client initialization to improve performance and reliability.

- **Refactor**
  - Renamed fields in various modules for better clarity (e.g., `K8s` to `K8sClient`, `Minio` to `MinioClient`).
  - Added a validation step in initialization functions to ensure proper configuration.

- **Bug Fixes**
  - Improved error handling with new error variables to address missing or mismatched configurations.

- **Tests**
  - Updated test cases to reflect changes in Minio and K8s client initialization.
  - Added new test scenarios for enhanced coverage.


pkg/minio/minio.go
```diff
@@ -44,6 +44,7 @@ const (
 type Minio struct {
 	Clientset kubernetes.Interface
 	Namespace string
+	cli       *miniogo.Client
```
MSevey, line 47: I'd call this `client`. I think `cli` is confusing since that is most often referred to for command line interface

pkg/minio/minio.go
```diff
@@ -160,26 +161,38 @@ func (m *Minio) IsMinioDeployed(ctx context.Context) (bool, error) {
...
 		return ErrMinioFailedToInitializeClient.Wrap(err)
 	}
 
-	if err := m.createBucketIfNotExists(ctx, cli, bucketName); err != nil {
+	return nil
+}
+
+// PushToMinio pushes data (i.e. a reader) to Minio
+func (m *Minio) PushToMinio(ctx context.Context, localReader io.Reader, minioFilePath, bucketName string) error {
+	if err := m.init(ctx); err != nil {
```
MSevey, line 187: If we are refactoring this, why not refactor to actually have a `New` method?  Instead of trying to, or checking for initialization on every call of this method?

MSevey (changes requested):
We should update this to have a `New` or `Init` method that is called once as opposed to requiring `init` to be called by every method.

pkg/minio/minio.go
mojtaba-esk: that makes sense, we can do the check in the knuu pkg and if user wants to use it, then we deploy it and start to use it

MSevey: Will you make these updates then? That was the original scope of the issue you made anyways. 
Or do you want to do that in a follow up?

mojtaba-esk: I will do them in this PR.

pkg/minio/minio.go
```diff
@@ -165,26 +166,38 @@ func (m *Minio) IsMinioDeployed(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
-// PushToMinio pushes data (i.e. a reader) to Minio
-func (m *Minio) PushToMinio(ctx context.Context, localReader io.Reader, minioFilePath, bucketName string) error {
+func (m *Minio) init(ctx context.Context) error {
+	if m.client != nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
mojtaba-esk: done

pkg/minio/minio.go
mojtaba-esk: It is done, could you please have a look?

pkg/minio/minio.go
```diff
@@ -51,7 +52,113 @@ type Config struct {
 	SecretAccessKey string
 }
 
-func (m *Minio) DeployMinio(ctx context.Context) error {
+func New(ctx context.Context, k8sClient k8s.KubeManager) (*Minio, error) {
+	m := &Minio{
+		K8s: k8sClient,
+	}
+
+	fmt.Println("minio.New is called")
```
MSevey, line 60: leftover println

mojtaba-esk: fixed it

e2e/tshark/tshark_test.go
```diff
@@ -48,7 +49,7 @@ func TestTshark(t *testing.T) {
 	require.NoError(t, err, "error setting command")
 
 	t.Log("deploying minio as s3 backend")
-	err = kn.MinioClient.DeployMinio(ctx)
+	kn.MinioClient, err = minio.New(ctx, kn.K8sClient)
```
MSevey, line 52: you shouldn't need this anymore right? because you passed in the `MinioEnabled: true` option. 
This would overwrite the minio client that was already created. 

If overwriting was the goal, we should add a comment explaining why we are doing that.

mojtaba-esk: fixed

pkg/knuu/knuu.go
```diff
@@ -42,9 +42,9 @@ type Options struct {
 	K8s          k8s.KubeManager
 	TestScope    string
 	ImageBuilder builder.Builder
-	Minio        *minio.Minio
-	Timeout      time.Duration
+	MinioEnabled bool
```
MSevey, line 45: I don't think we need the boolean. We can just have the `Minio` instance. 
Since it isn't a required service, developers can pass in `nil`. 
If they plan to use it, they just initialize the minio client and pass that in with the options.

pkg/knuu/knuu.go
```diff
@@ -69,6 +68,12 @@ func New(ctx context.Context, opts Options) (*Knuu, error) {
 		return nil, err
 	}
 
+	if opts.MinioEnabled {
+		if err := setupMinio(ctx, k); err != nil {
+			return nil, err
+		}
+	}
```
MSevey, line 75: Build on comment, this would get removed.

pkg/knuu/knuu.go
```diff
@@ -56,7 +56,6 @@ func New(ctx context.Context, opts Options) (*Knuu, error) {
 	k := &Knuu{
 		SystemDependencies: system.SystemDependencies{
 			K8sClient:    opts.K8s,
-			MinioClient:  opts.Minio,
```
MSevey, line 59: Building on comment, we would revert this.

pkg/knuu/knuu.go
```diff
@@ -193,16 +198,9 @@ func setDefaults(ctx context.Context, k *Knuu) error {
 		}
 	}
 
-	if k.MinioClient == nil {
-		k.MinioClient = &minio.Minio{
-			K8s: k.K8sClient,
-		}
-	}
-
```
MSevey, line 201: Building on comment, we would still remove this in order to make defining the `MinioClient` optional.

pkg/knuu/knuu.go
```diff
@@ -223,3 +221,8 @@ func setupProxy(ctx context.Context, k *Knuu) error {
 	k.Logger.Debugf("Proxy endpoint: %s", endpoint)
 	return nil
 }
+
+func setupMinio(ctx context.Context, k *Knuu) (err error) {
+	k.MinioClient, err = minio.New(ctx, k.K8sClient)
+	return err
+}
```
MSevey, line 228: Building on comment, we wouldn't need this. This would become part of test helpers or something.

pkg/knuu/knuu.go
mojtaba-esk: Ideally we could do it, then we need to also ask user to pass k8sClient as well since we need it for deploying minio.
Also we might also question the enabling proxy which is also a boolean.
So wdyt?

mojtaba-esk: I think also receiving k8s client from user makes more sense

MSevey: The k8 client is already passed in by the user, so the user has that to initialize the minio client. 

It also causes less of a diff, since there isn't a new option. It is just the change in default behavior that is impacted. 

For how the `New` method is currently structured, I'd recommend accepting the actual clients vs flags. I wouldn't change the `ProxyEnabled` flag just yet though. We should review the New method again after this PR is merged and see if there are any other updates we want to make as part of the breaking release.

mojtaba-esk: Ok pushed some changes. k8sClient was accepted as an option and if it was not passed, a default client would have been initiated. Now, user has to pass it as well. I think now it is in a pretty good shape. thanks for pointing it out.
Agreed that the proxy is something more embedded in knuu and refactoring its option requires more thought.

pkg/knuu/knuu.go
```diff
@@ -39,24 +39,23 @@ type Knuu struct {
...
-	Minio        *minio.Minio
-	Timeout      time.Duration
+	MinioClient  *minio.Minio
 	ProxyEnabled bool
+	Timeout      time.Duration
 	Logger       *logrus.Logger
 }
 
-func New(ctx context.Context, opts Options) (*Knuu, error) {
+func New(ctx context.Context, k8sClient k8s.KubeManager, opts Options) (*Knuu, error) {
```
MSevey, line 50: why did you pull this out of the options? we should just leave it in the options.

pkg/knuu/knuu.go
```diff
@@ -171,38 +175,22 @@ func loadEnvVariables() error {
...
 		k.timeout = defaultTimeout
 	}
 
-	if k.K8sClient == nil {
-		var err error
-		k.K8sClient, err = k8s.NewClient(ctx, k.TestScope)
-		if err != nil {
-			return ErrCannotInitializeK8s.Wrap(err)
-		}
-	}
```
MSevey, line 194: this should stay in.

pkg/knuu/knuu.go
```diff
@@ -82,6 +81,11 @@ func New(ctx context.Context, opts Options) (*Knuu, error) {
 	return k, nil
 }
 
+func DefaultTestScope() string {
+	t := time.Now()
+	return fmt.Sprintf("%s-%03d", t.Format("20060102-150405"), t.Nanosecond()/1e6)
+}
+
```
MSevey, line 88: We shouldn't put test helpers in production code. 
Test helpers can be put in the `e2e` package.

e2e/bittwister/suite_setup_test.go
```diff
@@ -16,11 +17,11 @@ type Suite struct {
...
 
 func (s *Suite) SetupSuite() {
-	var (
-		err error
-		ctx = context.Background()
-	)
-	s.Knuu, err = knuu.New(ctx, knuu.Options{ProxyEnabled: true})
+	ctx := context.Background()
+	k8sClient, err := k8s.NewClient(ctx, knuu.DefaultTestScope())
+	s.Require().NoError(err)
```
MSevey, line 22: if we revert your change of adding the k8 client as a top level argument and removing it from the `setdefaults` function, then you could revert this.

e2e/system/build_from_git_test.go
```diff
@@ -18,7 +19,11 @@ func TestBuildFromGit(t *testing.T) {
 	ctx := context.Background()
 
 	// The default image builder is kaniko here
-	kn, err := knuu.New(ctx, knuu.Options{})
+
+	k8sClient, err := k8s.NewClient(ctx, knuu.DefaultTestScope())
+	require.NoError(t, err, "error creating k8s client")
+
+	kn, err := knuu.New(ctx, k8sClient, knuu.Options{})
```
MSevey, line 26: same here, as to why we shouldn't have the k8client as a standalone argument.

pkg/knuu/knuu.go
mojtaba-esk: because this is mandatory to have and it is not an optional parameter.
If we want to have Minio and probably some other services in future initiated by the user, which is a good practice, then we need to  use the same `k8sClient` for both. For example if we use one `k8sClient` for Minio and `Knuu` also makes its own, we will have two `k8sClinets` which introduces inconsistency also it might point to a different namespaces. So we have to extract it and ask user to pass it in.

pkg/knuu/knuu.go
mojtaba-esk: agreed

MSevey (changes requested):
final clean up

e2e/system/build_from_git_test.go
```diff
@@ -18,7 +20,11 @@ func TestBuildFromGit(t *testing.T) {
 	ctx := context.Background()
 
 	// The default image builder is kaniko here
-	kn, err := knuu.New(ctx, knuu.Options{})
+
+	k8sClient, err := k8s.NewClient(ctx, e2e.DefaultTestScope())
+	require.NoError(t, err, "error creating k8s client")
+
+	kn, err := knuu.New(ctx, knuu.Options{K8sClient: k8sClient})
```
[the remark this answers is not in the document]
mojtaba-esk: that's true. ill clean it up asap

pkg/knuu/knuu.go
```diff
@@ -176,9 +193,8 @@ func setDefaults(ctx context.Context, k *Knuu) error {
 		k.Logger = log.DefaultLogger()
 	}
 
-	if k.TestScope == "" {
-		t := time.Now()
-		k.TestScope = fmt.Sprintf("%s-%03d", t.Format("20060102-150405"), t.Nanosecond()/1e6)
+	if k.TestScope == "" && k.K8sClient != nil {
+		k.TestScope = k.K8sClient.Namespace()
```
MSevey, line 197: You're missing the default base case of setting TestScope when both are empty. 
```suggestion
	if k.TestScope == "" && k.K8sClient != nil {
		k.TestScope = k.K8sClient.Namespace()
	} else if k.TestScope == "" {
		t := time.Now()
		k.TestScope = fmt.Sprintf("%s-%03d", t.Format("20060102-150405"), t.Nanosecond()/1e6)
```

pkg/knuu/knuu_test.go
```diff
@@ -66,70 +66,84 @@ func TestNew(t *testing.T) {
 	}{
 		{
 			name:        "Default initialization",
-			options:     Options{},
+			options:     Options{TestScope: "test"},
```
MSevey, line 69: per my original comment, this case can now be reverted as `TestScope` shouldn't be required.

pkg/knuu/knuu_test.go
```diff
@@ -66,70 +66,84 @@ func TestNew(t *testing.T) {
...
-				assert.NotNil(t, k.MinioClient)
+				assert.NotNil(t, k.ImageBuilder)
 			},
 		},
 		{
-			name: "With custom Image Builder",
+			name:        "Without TestScope and K8sClient",
+			options:     Options{},
+			expectError: true,
+		},
```
MSevey, line 137: this should then be removed. as this is the base case

pkg/knuu/knuu_test.go
```diff
@@ -147,3 +161,51 @@ func TestNew(t *testing.T) {
...
+				TestScope: "",
+				K8sClient: nil,
+			},
+			wantErr: true,
+		},
+		{
+			name: "Valid options",
+			options: Options{
+				TestScope: "test",
+			},
```
MSevey, line 198: this should be default empty options.

pkg/knuu/knuu_test.go
```diff
@@ -147,3 +161,51 @@ func TestNew(t *testing.T) {
...
+			wantErr: false,
+		},
+		{
+			name: "TestScope and K8sClient not set",
+			options: Options{
+				TestScope: "",
+				K8sClient: nil,
+			},
+			wantErr: true,
```
MSevey, line 192: this can be removed as this is just the base case that should be valid.

pkg/knuu/knuu.go
```diff
@@ -160,6 +164,19 @@ func (k *Knuu) handleTimeout(ctx context.Context) error {
...
+func validateOptions(opts Options) error {
+	// When Minio is set, K8sClient must be set too
+	// to make sure that there is only one source of truth for the k8s client
+	if opts.MinioClient != nil && opts.K8sClient == nil {
+		return ErrK8sClientNotSet
+	}
+
+	if opts.TestScope == "" && opts.K8sClient == nil {
+		return ErrTestScopeNotSet
+	}
```
MSevey, line 176: This isn't actually a problem. Now that you've moved the validation logic out you can see that this is simply the base case. You should be able to not set these. 

The case you want to check is this:
```suggestion
	if opts.TestScope != "" && opts.K8sClient != nil && opts.TestScope != opts.K8sClient.Namespace() {
		return ErrTestScopeMistMatch
	}
```

e2e/system/build_from_git_test.go
```diff
@@ -18,7 +23,66 @@ func TestBuildFromGit(t *testing.T) {
 	ctx := context.Background()
 
 	// The default image builder is kaniko here
-	kn, err := knuu.New(ctx, knuu.Options{})
+	kn, err := knuu.New(ctx, knuu.Options{TestScope: e2e.DefaultTestScope()})
```
MSevey, line 26: fixing the validation logic will revert this.

e2e/system/build_from_git_test.go
```diff
@@ -18,7 +23,66 @@ func TestBuildFromGit(t *testing.T) {
...
+
+	// The file is created by the dockerfile in the repo,
+	// so to make sure it is built correctly, we check the file
+	data, err := target.GetFileBytes(ctx, "/test.txt")
+	require.NoError(t, err, "Error getting file bytes")
+
+	data = []byte(strings.TrimSpace(string(data)))
+	assert.Equal(t, []byte("Hello, World!"), data, "File bytes do not match")
+}
+func TestBuildFromGitWithModifications(t *testing.T) {
```
MSevey, line 67: Why did we need to add this test to this PR?

pkg/knuu/knuu_test.go
```diff
@@ -66,70 +66,84 @@ func TestNew(t *testing.T) {
...
-				assert.NotNil(t, k.MinioClient)
 				assert.NotNil(t, k.ImageBuilder)
 				assert.Equal(t, defaultTimeout, k.timeout)
 			},
 		},
 		{
-			name: "With custom Logger",
+			name:        "With custom Minio client without setting k8sClient",
+			options:     Options{MinioClient: &minio.Minio{}},
+			expectError: true,
```
MSevey, line 82: To help catch some of the logic errors we are uncovering in this PR, I'd recommend you update this test struct to accept the actual error, not just whether you expect an error.

pkg/knuu/knuu_test.go
```diff
@@ -147,3 +161,51 @@ func TestNew(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateOptions(t *testing.T) {
+	tests := []struct {
+		name    string
+		options Options
+		wantErr bool
```
MSevey, line 169: this should be `err` and then we set the actual error we expect. This helps enough we are actually hitting the expected error and not just any error.

pkg/knuu/knuu_test.go
MSevey: the test case should be validating that when they are both set but don't match, it errors.

pkg/knuu/knuu.go
mojtaba-esk: well it was like this before, but in one of the comments you told me to move this default test scope to the e2e pkg: 
https://github.com/celestiaorg/knuu/pull/434/files/41fff2e769f1e40f1a336d265a1579265647fc47#diff-04a1accd1c12dafffe8d3ecc43a623cd0bf22d03d3561f7e04e296e64962048eR8-R11
that's the reason for receiving it from the user. I am more fan of making it easier for the users, so will put it back.

pkg/knuu/knuu.go
mojtaba-esk: well it was like this before, but in one of the comments you told me to move this default test scope to the e2e pkg: 
https://github.com/celestiaorg/knuu/pull/434/files/41fff2e769f1e40f1a336d265a1579265647fc47#diff-04a1accd1c12dafffe8d3ecc43a623cd0bf22d03d3561f7e04e296e64962048eR8-R11
that's the reason for receiving it from the user. I am more fan of making it easier for the users, so will put it back.

e2e/system/build_from_git_test.go
mojtaba-esk: because we refactored how user enables Minio and Minio is required by the builder when builds on directory context and since this test modifies the built image, it does it in a tmp directory which is local.

pkg/knuu/knuu.go
MSevey: Sorry, the name of this variable and the function is confusing. 
When it was refactored it looked like a testing variable, which is why I made the original comment. 
Knuu is a Test framework, which makes sense why this is called TestScope. I see that now. 

Maybe in a follow up we could rename this to just be Scope to avoid future confusion

e2e/system/build_from_git_test.go
MSevey: ah ok, thanks.

pkg/knuu/knuu.go
mojtaba-esk: yeah that's a good idea to rename it.

e2e/system/build_from_git_test.go
```diff
@@ -21,6 +25,65 @@ func TestBuildFromGit(t *testing.T) {
 	kn, err := knuu.New(ctx, knuu.Options{})
 	require.NoError(t, err, "Error creating knuu")
 
+	target, err := kn.NewInstance("git-builder")
```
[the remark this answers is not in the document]
mojtaba-esk: it is done in `knuu.New`

MSevey (approved):
Nice!

Issue #410, opened 2024-05-31
Knuu Options Struct


MSevey:
## Current

With the new refactor the following knuu option structure was introduced. 

```go
type Option func(*Knuu)

func WithImageBuilder(builder builder.Builder) Option {
	return func(k *Knuu) {
		k.ImageBuilder = builder
	}
}
```

The goal was to provide flexibility for developers to enable what they wanted. 

Two downsides to this approach are:
1. Every new option requires a new function
2. Every new option requires a function call when call knuu.New()

Additionally this has the New method applying options and then setting defaults.  
This approach is flexible, but could lead to developer errors because the method is making assumptions based on inputs verse handling those inputs more strictly. 

## Proposed Change

If we use an Options struct we can have the same flexibility for developers with the following benefits:
1. No new function definitions
2. No additional function calls on knuu.New()
4. Ability to define DefaultOptions

To start the Options struct would be the following.

```go
type Options struct {
	system.SystemDependencies
	timeout      time.Duration
	proxyEnabled bool
}
```

This currently is basically duplicating the Knuu struct, but the benefit here is that it allows us to update the knuu and options structs independently. 

This would also allow us to update the New method in a way that still provides flexibility while also being more strict on validating inputs. 

```go
func New(ctx context.Context) (*Knuu, error) {
  return NewWithOptions(ctx, DefaultOptions())
}

func NewWithOptions(ctx context.Context, opts Options) (*Knuu, error) {
  // First validate options
  ...  
  // Then apply
  k := &Knuu{opts}
  // Do other things
  return k, nil
}

func DefaultOptions() Options {
  // Initialize defaults
  return
}
```

This allows the use of `New()` without needing to worrying about any options, which is create to testing and getting started. 
Also refactoring out the `DefaultOptions` makes it easy for developers to get the defaults and just update the options they need to.

PR #436, opened 2024-06-10
chore: refactor knuu object to use options as argument

mojtaba-esk:
This PR proposes a change to use options as an argument in the new function instead of using WithFuncs. 

## Summary by CodeRabbit

- **Refactor**
  - Improved code readability and maintainability by centralizing configuration options into a single `Options` struct.
  - Replaced individual option functions with a unified `Options` struct for easier management.

- **Tests**
  - Updated test cases to use the new `Options` struct instead of individual options, ensuring consistency and simplifying test setup.

### DIFF
--- a/pkg/knuu/knuu.go
+++ b/pkg/knuu/knuu.go
@@ -166,7 +166,7 @@ func loadEnvVariables() error {
 		return ErrCannotLoadEnv.Wrap(err)
 	}
 	if os.IsNotExist(err) {
-	        logrus.Info("The .env file does not exist, continuing without loading environment variables.")
+		logrus.Info("The .env file does not exist, continuing without loading environment variables.")
 	}
 	return nil
 }
--- a/pkg/knuu/knuu.go
+++ b/pkg/knuu/knuu.go
@@ -161,10 +161,13 @@ func (k *Knuu) handleTimeout(ctx context.Context) error {
 }
 
 func loadEnvVariables() error {
-	if err := godotenv.Load(); err != nil && !os.IsNotExist(err) {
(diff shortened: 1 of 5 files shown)


pkg/knuu/knuu.go
```diff
@@ -236,3 +159,66 @@ func (k *Knuu) handleTimeout(ctx context.Context) error {
 
 	return nil
 }
+
+func loadEnvVariables() error {
+	if err := godotenv.Load(); err != nil && !os.IsNotExist(err) {
+		return ErrCannotLoadEnv.Wrap(err)
+	}
+	logrus.Info("The .env file does not exist, continuing without loading environment variables.")
+	return nil
```
MSevey, line 168: won't this always log that the env file doesn't exist even if it does? 
Shouldn't we update to:
```suggestion
	err := godotenv.Load()
	if err != nil && !os.IsNotExist(err) {
		return ErrCannotLoadEnv.Wrap(err)
	}
	if os.IsNotExist(err) {
	        logrus.Info("The .env file does not exist, continuing without loading environment variables.")
	}
	return nil
```

MSevey (dismissed):
Minor comment on the env loading. Fine to do as a follow up though. 

Also let's reference the issue this is closing to make sure it closes it on merge.

pkg/knuu/knuu.go
mojtaba-esk: Nice catch!

MSevey (approved):
Closes #410

mojtaba-esk:
> Closes #410

Thanks for finding it. I did not.

PR #437, opened 2024-06-10
refactor: change basic test to suite

sysrex:
Change tests to test suite.

## Summary by CodeRabbit

- **Refactor**
  - Improved test logic and structure for better instance management and context handling.
  - Introduced a `TestSuite` struct for organized test setup and cleanup.

- **Style**
  - Adjusted import order in `main_test.go` for better readability and organization.

- **New Features**
  - Added a new test suite setup file `suite_setup_test.go` to streamline the testing process.


e2e/basic/basic_test.go
```diff
@@ -1,53 +1,62 @@
...
+			s.T().Logf("error destroying instances: %v", err)
+		}
 	})
 
-	// Test logic
+	// Test Logic
+	s.Require().NoError(target.Start(ctx))
+	s.Require().NoError(target.WaitInstanceIsRunning(ctx))
+
+	s.Require().NoError(err)
```
MSevey, line 39: I think this is not needed. There is no new error here.

e2e/basic/basic_test.go
```diff
@@ -1,53 +1,62 @@
...
+	type testCase struct {
+		name string
 	}
-	err = instance.WaitInstanceIsRunning()
-	if err != nil {
-		t.Fatalf("Error waiting for instance to be running: %v", err)
+
+	tt := []testCase{
+		{"Test Case 1"},
+		{"Test Case 2"},
```
MSevey, line 48: For this test, we really just need 1 test case, so we can just directly call what is on lines 55 to 58

MSevey:
golangci-lint errors can be fixed by running `make fmt`

e2e/basic/suite_setup_test.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (s *Suite) SetupSuite() {
+	var (
+		err error
+		ctx = context.Background()
+	)
+	s.Knuu, err = knuu.New(ctx)
+	s.Require().NoError(err)
+	s.FailNow("Error cleaning up test suite: %v", err)
```
MSevey, line 25: This is failing all tests on this line. Needs to be removed. 
```suggestion
```

e2e/basic/suite_setup_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+type Suite struct {
```
MSevey, line 13: We should update this to be a unique name from `suite.Suite` to avoid errors from import conflicts. 
```suggestion
type TestSuite struct {
```
This change would need to be reflected in the methods in this file and the test you updated

e2e/basic/suite_setup_test.go
```diff
@@ -0,0 +1,41 @@
...
+	s.Knuu.HandleStopSignal(ctx)
+}
+
+func (s *Suite) TearDownSuite() {
+	s.T().Cleanup(func() {
+		logrus.Info("Tearing down test suite...")
+		err := s.Knuu.CleanUp(context.Background())
+		if err != nil {
+			s.FailNow("Error cleaning up test suite: %v", err)
+		}
```
MSevey, line 35: You can simplify this:
```suggestion
		s.Require().NoError(s.Knuu.CleanUp(context.Background()))
```

e2e/basic/suite_setup_test.go
MSevey: Here we use it as TestSuite as well https://github.com/celestiaorg/knuu/pull/416/files#diff-f45d80f83b805ff61cdd3c3aa3a3de08045cdc644649e042af4ff3a786f78855R16

e2e/basic/main_test.go
```diff
@@ -2,19 +2,29 @@ package basic
...
+
+	// Perform the test
+	exitVal := s.Run("TestMain", func() {
+		s.T().Logf("Scope: %s", s.Knuu.Scope())
+	})
+
+	exitValue := convertViaMap(exitVal)
+
+	os.Exit(exitValue)
 }
```
mojtaba-esk, line 30: This testmain func is not really needed. if we convert all our basic tests to test suit, this func can be removed

e2e/basic/basic_test.go
```diff
@@ -1,53 +1,60 @@
...
-	"github.com/stretchr/testify/require"
 
-	"github.com/celestiaorg/knuu/pkg/knuu"
+	"github.com/celestiaorg/knuu/pkg/instance"
 )
 
-func TestBasic(t *testing.T) {
-	t.Parallel()
+const (
+	TestImage = "alpine:latest"
```
mojtaba-esk, line 12: two points about this one:
1. This does not have to be starting with a Capital letter as we do not really need to export this constant.
2. It can have some more specific name as it is accessible package wise so other basic tests also can see this.

e2e/basic/basic_test.go
```diff
@@ -1,53 +1,60 @@
...
+		{"Hello World"},
 	}
-	wget, err := instance.ExecuteCommand("echo", "Hello World!")
-	if err != nil {
-		t.Fatalf("Error executing command '%v':", err)
+
+	for _, tc := range tt {
+		tc := tc
+		s.Run(tc.name, func() {
+			s.T().Logf("Running test case: %s", tc.name)
```
mojtaba-esk, line 52: no need to add this line as the Run func prints the test name itself.

e2e/basic/suite_setup_test.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (s *TestSuite) SetupSuite() {
+	var (
+		err error
+		ctx = context.Background()
+	)
+	s.Knuu, err = knuu.New(ctx)
+	s.Require().NoError(err)
+	s.T().Logf("Error cleaning up test suite: %v", err)
```
mojtaba-esk, line 25: This line can be removed as it seems like it is a typo to be here

e2e/basic/suite_setup_test.go
```diff
@@ -0,0 +1,38 @@
...
+	s.Knuu.HandleStopSignal(ctx)
+}
+
+func (s *TestSuite) TearDownSuite() {
+	s.T().Cleanup(func() {
+		logrus.Info("Tearing down test suite...")
+		s.Require().NoError(s.Knuu.CleanUp(context.Background()))
```
mojtaba-esk, line 32: Let's use just the test Log instead of Require so if the cleanup fails, the test wont.

mojtaba-esk (approved):
Great job. Overall everything went very well. Left some suggestion comments.

smuu (approved):

MSevey:
@coderabbitai resolve

e2e/basic/basic_test.go
```diff
@@ -1,53 +1,59 @@
...
+	type testCase struct {
+		name string
 	}
-	err = instance.WaitInstanceIsRunning()
-	if err != nil {
-		t.Fatalf("Error waiting for instance to be running: %v", err)
+
+	tt := []testCase{
+		{"Hello World"},
 	}
```
MSevey, line 47: You don't need to define this struct here when you just have a single test case.
```suggestion
```

e2e/basic/basic_test.go
```diff
@@ -1,53 +1,59 @@
...
-		t.Fatalf("Error executing command '%v':", err)
+
+	for _, tc := range tt {
+		tc := tc
+		ts.Run(tc.name, func() {
+			output, err := target.ExecuteCommand(ctx, "echo", tc.name)
+			ts.Require().NoError(err)
+
+			assert.Contains(ts.T(), output, tc.name)
+		})
```
MSevey, line 56: you don't need a for loop if you just have a single test case. just test it directly.
```suggestion
                input := "Hello World"
		ts.Run(tc.name, func() {
			output, err := target.ExecuteCommand(ctx, "echo", input)
			ts.Require().NoError(err)

			assert.Contains(ts.T(), output, input)
		})
```

e2e/basic/main_test.go
```diff
@@ -2,19 +2,29 @@ package basic
...
+func convertViaMap(b bool) int {
+	table := map[bool]int{
+		true:  1,
+		false: 0,
 	}
-	logrus.Infof("Scope: %s", knuu.Scope())
-	exitVal := m.Run()
-	os.Exit(exitVal)
+	return table[b]
+}
```
MSevey, line 13: For something like this, it isn't worth allocating a map.
It is easier to just have a single if condition.
```go
func boolToExitCode(b bool) int {
        if b {
                return 1
        }
        return 0
}
```

MSevey (changes requested):

e2e/basic/main_test.go
MSevey: also I'm not sure why this is even needed? 
why are we using `os.Exit` in a test? 
Why aren't we just using `ts.Require().True(exitVal)` if that is what we are looking for?

e2e/basic/basic_test.go
smuu: prep for having multiple tests

e2e/basic/basic_test.go
smuu: prep for having multiple tests

e2e/basic/main_test.go
smuu: will be removed in another PR

e2e/basic/main_test.go
smuu: main will be deleted in another PR, we use test suite

PR #443, opened 2024-06-19
refactor!: sidecars to inetrface

### LINKED ISSUE #431: Refactor sidecars to an inetrface
Currently we have a multiple predefined sidecars that user can enable including: `OtelCollector`, `TsharkCollector`, `BitTwister`, and there might be more in future.
to have a clean code base with a nice structure, let's define the sidecar as an interface that each new sidecar implements it. it enhances also the composeability.

mojtaba-esk:
Closes #431 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `NetShaper` component to manage network shaping configurations and interactions with the BitTwister sidecar, enhancing performance control.
	- Implemented comprehensive test suites for the `NetShaper`, `Observability`, and `ReverseProxy` components, improving validation and reliability.
	- Enhanced observability features with OpenTelemetry integration for better monitoring and tracing capabilities.

- **Bug Fixes**
	- Improved error handling with specific messages for various failure scenarios related to the BitTwister service and observability components.

- **Documentation**
	- Added documentation for the new `NetShaper` functionalities, observability features, and error types to facilitate user understanding.

- **Refactor**
	- Enhanced the architecture of tests and components for better encapsulation and maintainability while streamlining the `Instance` management process.


pkg/knuu/instance_old.go
```diff
@@ -199,48 +199,8 @@ func (i *Instance) SetStartupProbe(startupProbe *v1.Probe) error {
...
-
-// Deprecated: Use the new package knuu instead.
-func (i *Instance) SetPrometheusExporter(endpoint string) error {
-	return i.Instance.SetPrometheusExporter(endpoint)
-}
-
-// Deprecated: Use the new package knuu instead.
-func (i *Instance) SetPrometheusRemoteWriteExporter(endpoint string) error {
-	return i.Instance.SetPrometheusRemoteWriteExporter(endpoint)
+func (i *Instance) AddSidecar(ctx context.Context, sc instance.SidecarManager) error {
```
smuu, line 202: It would be great to override the default image used in the sidecar if you want to test another version.

mojtaba-esk: When user creates the sidecar object, they can assign any image they want, if not the default image will be considered. Is that ok?

pkg/sidecars/tshark/tshark_test.go
mojtaba-esk, line 1: This file has to be deleted

pkg/knuu/instance_old.go
mojtaba-esk: Just double checked it, for some sidecars we do have it. Let me modify the others code so we can have customized image if user wants.

mojtaba-esk: a fix added

e2e/tshark/tshark_test.go
```diff
@@ -41,58 +40,42 @@ func TestTshark(t *testing.T) {
...
 	t.Log("deploying minio as s3 backend")
-	err = kn.MinioClient.DeployMinio(ctx)
-	require.NoError(t, err, "error deploying minio")
+	require.NoError(t, kn.MinioClient.DeployMinio(ctx))
 
 	t.Log("getting minio configs")
 	minioConf, err := kn.MinioClient.GetConfigs(ctx)
 	require.NoError(t, err, "error getting S3 (minio) configs")
 
+	keyPrefix := "tshark/" + scope
```
smuu, line 53: We have `obsy.New()` and `bittwister.New()`. Why we don't have `tshark.New()` here?

pkg/instance/instance.go
```diff
@@ -867,181 +785,32 @@ func (i *Instance) SetStartupProbe(startupProbe *v1.Probe) error {
 
 // AddSidecar adds a sidecar to the instance
 // This function can only be called in the state 'Preparing' or 'Committed'
-func (i *Instance) AddSidecar(sidecar *Instance) error {
-
-	if !i.IsInState(Preparing, Committed) {
-		return ErrAddingSidecarNotAllowed.WithParams(i.state.String())
-	}
-	if sidecar == nil {
+func (i *Instance) AddSidecar(ctx context.Context, sc SidecarManager) error {
```
smuu, line 788: Now, the dev needs to understand what a Kubernetes sidecar is when adding obsy, bittwister, or tshark. The knowledge we assume from the devs should be as minimal as possible. Before the dev was just adding the features and under the hood, knuu added these as a sidecar.
Any ideas on how we can retain that property?

pkg/sidecars/bittwister/bittwister.go
smuu, line 1: Do we want to rename bittwister to something like 'networkShaping' so that it is obvious for new devs?

pkg/sidecars/tshark/helpers.go
smuu, line 1: `SetImage(image string)` is missing for tshark.

e2e/tshark/tshark_test.go
mojtaba-esk: Because they have some initilization like assigning default values, but tshark does not have. We can have it for sake of consistency though

pkg/instance/instance.go
mojtaba-esk: Maybe we can call it a different name as it is kinda pluggin for knuu and users can develop their own sidecars if they want. However, I also think that users do not need to know alot about sidecars, maybe what they understand from knuu sidecar is much more simplified than k8s sidecar as there is a sidecar interface with a couple of funcs that they need to implement. I am ok with calling it a different name. Do you have any proposal how to make it more friendly to the users, not only renaming it, in general?

smuu: I think moving these functions to a separate object is a significant improvement to the status quo.

I came up with these alternative names:
- Capability
- Feature
- Ability
- Module

But maybe it is fine to keep the sidecar term, as you said, the knuu sidecar is much more simplified than the k8s sidecar.

Would like to have input from @tty47

pkg/sidecars/tshark/helpers.go
mojtaba-esk: User can set it directly. However, if we add a New func, we can unexport the image and let user to set it if they are not happy with the default

smuu: bittwister and obsy have this function. It would be nice to have it consistent to set another image.

e2e/tshark/tshark_test.go
mojtaba-esk: oh, I just saw that the tShark struct has too many params and does not really need an initialization. see:

```go
type Tshark struct {
	Image    string
	VolumeSize resource.Quantity
	S3AccessKey string
	S3SecretKey string
	S3Region string
	S3Bucket string
	CreateBucket bool
	S3KeyPrefix string
	S3Endpoint string
	UploadInterval time.Duration
}
```

so if we use a New func, it will need to pass an option which has the whole fields and it will be just redundant. So I suggest to keep it as it is right now.

pkg/sidecars/tshark/tshark_test.go
mojtaba-esk: Unittests added to it

pkg/instance/instance.go
mojtaba-esk: I like the name `Module` more than others

pkg/sidecars/bittwister/bittwister.go
mojtaba-esk: BitTwister has a separate repo. We can rename the sidecar to some `netwokrShaping` name to have an abstract layer on it. I am ok with either keeping it this name or change it to a new name. If we keep the name, we need to properly explain it in the docs so users know what it is and what it does.

pkg/instance/instance.go
mojtaba-esk: mm I love the `Plugin` name more.

pkg/sidecars/bittwister/bittwister.go
mojtaba-esk: ok will do it

pkg/instance/instance.go
```diff
@@ -515,9 +425,9 @@ func (i *Instance) copyFileToBuildDir(src, dest string) (string, error) {
 }
 
 func (i *Instance) addFileToInstance(dstPath, dest, chown string) error {
-	if !i.isSubFolderOfVolumes(dest) {
-		return ErrFileIsNotSubFolderOfVolumes.WithParams(dest)
-	}
+	// if !i.isSubFolderOfVolumes(dest) {
```
mojtaba-esk, line 428: @smuu do you know why these lines are commented out?

pkg/sidecars/bittwister/bittwister.go
mojtaba-esk: done

smuu (approved):

PR #445, opened 2024-06-19
chore: refactor to use logger obj instead

### LINKED ISSUE #396: Consider refactoring all pkgs to receive a logger object instead of default logrus
Receiving a logger object from user makes our packages more composable. Right now only `knuu` pkg does this, this issue suggest to propagate it throughout the entire repo.

mojtaba-esk:
Closes #396 

## Summary by CodeRabbit

- **Refactor**
  - Replaced the `logrus` logging mechanism with a custom logger (`i.Logger` or `c.logger`) across various modules for enhanced logging control and flexibility.
  - Updated function signatures to include the custom logger as an argument.

- **Chores**
  - Removed unnecessary imports of `logrus` from multiple files.

These changes improve the overall logging consistency and maintainability within the codebase.


pkg/instance/helper.go
```diff
@@ -332,13 +330,13 @@ func (i *Instance) destroyResources(ctx context.Context) error {
 		// enable network when network is disabled
 		disableNetwork, err := i.NetworkIsDisabled(ctx)
 		if err != nil {
-			logrus.Debugf("error checking network status for instance")
+			i.Logger.Debugf("error checking network status for instance")
```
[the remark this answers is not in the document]
MSevey: agree, worth updating

pkg/k8s/k8s_namespace.go
```diff
@@ -47,7 +46,7 @@ func (c *Client) GetNamespace(ctx context.Context, name string) (*corev1.Namespa
 func (c *Client) NamespaceExists(ctx context.Context, name string) bool {
 	_, err := c.GetNamespace(ctx, name)
 	if err != nil {
-		logrus.Debugf("Namespace %s does not exist, err: %v", name, err)
+		c.logger.Debugf("Namespace %s does not exist, err: %v", name, err)
```
MSevey, line 49: Similar comment, might as well make this `Errorf`

pkg/k8s/k8s_networkpolicy.go
```diff
@@ -92,7 +91,7 @@ func (c *Client) GetNetworkPolicy(ctx context.Context, name string) (*v1.Network
 func (c *Client) NetworkPolicyExists(ctx context.Context, name string) bool {
 	_, err := c.GetNetworkPolicy(ctx, name)
 	if err != nil {
-		logrus.Debug("NetworkPolicy does not exist, err: ", err)
+		c.logger.Debug("NetworkPolicy does not exist, err: ", err)
```
MSevey, line 94: same

pkg/knuu/knuu.go
```diff
@@ -69,6 +65,10 @@ func New(ctx context.Context, opts Options) (*Knuu, error) {
 		return nil, err
 	}
 
+	if err := k.loadEnvVariables(); err != nil {
```
MSevey, line 68: hmmm, this is a challenge I think.  Currently the `setDefaults` doesn't rely on any env vars but that could change in the future. 

I think it is ok to leave the `loadEnvVariables` at the top of the function and uses the default logrus logger for the one log line about the `.env` file so that we don't create a bigger issue in the future. 

We could simply add a comment explaining why we are leaving it the way we are so that we don't try and update in the future.

MSevey (changes requested):
quick fixes.

pkg/k8s/k8s_namespace.go
mojtaba-esk: This is usually not an error, it just tells that if the namespace exist or not.  Although we need to check if the type of the returned error is `NotFound`.
will fix it

MSevey (changes requested):
One unresolved comment from previous review

pkg/knuu/knuu.go
mojtaba-esk: Currently knuu options are set via its `New` func. The only thing is the default logger which reads the log level from env var. I think that loading `.env` file is more for user's usage. honestly I am not a big fan of it to be called with the `New` func. This is something that User had to take care of.

mojtaba-esk: I'll move to top anyways. created an issue to handle it #457

MSevey (approved):

PR #465, opened 2024-06-28
fix: sanitize scope

smuu:
## Overview

When initializing knuu with an upper case letter, the `validateOptions` function fails as the scope name is not sanitized correctly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the creation process of new `Knuu` instances by sanitizing the `TestScope` field to ensure it meets validation criteria, enhancing stability and reliability.

### DIFF
--- a/pkg/knuu/knuu.go
+++ b/pkg/knuu/knuu.go
@@ -49,6 +49,8 @@ type Options struct {
 }
 
 func New(ctx context.Context, opts Options) (*Knuu, error) {
+	opts.TestScope = k8s.SanitizeName(opts.TestScope)
+
 	if err := validateOptions(opts); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/knuu/knuu.go
```diff
@@ -49,6 +49,8 @@ type Options struct {
 }
 
 func New(ctx context.Context, opts Options) (*Knuu, error) {
+	opts.TestScope = k8s.SanitizeName(opts.TestScope)
```
mojtaba-esk, line 52: if the testscope is empty, the SanitizeName will panic, but the expected behaviour is to fill it with the default scope.

mojtaba-esk (approved):

pkg/knuu/knuu.go
mojtaba-esk, line 52: if the testscope is empty, the SanitizeName will panic, but the expected behaviour is to fill it with the default scope.

mojtaba-esk (changes requested):

PR #497, opened 2024-07-08
fix: otel sidecar

smuu:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of errors related to adding a volume for the OpenTelemetry (otel) agent instance to prevent configuration issues.

- **Updates**
  - Enhanced OpenTelemetry agent configuration by adding a volume, updating file paths, and adjusting command settings to ensure better resource management and configuration accuracy.


mojtaba-esk (approved):
Thanks for the fix.
we could also extract those constants.

PR #502, opened 2024-07-15
chore: Deploy timeout-handler before traefik and minio

### LINKED ISSUE #462: Deploy timeout-handler before traefik and minio

mojtaba-esk:
Closes #462 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling with the addition of a new timeout error message.
  - Enhanced initialization process for better performance and reliability.

### DIFF
--- a/pkg/knuu/knuu.go
+++ b/pkg/knuu/knuu.go
@@ -49,8 +49,6 @@ type Options struct {
 }
 
 func New(ctx context.Context, opts Options) (*Knuu, error) {
-	opts.TestScope = k8s.SanitizeName(opts.TestScope)
-
 	if err := validateOptions(opts); err != nil {
 		return nil, err
 	}
@@ -79,10 +77,6 @@ func New(ctx context.Context, opts Options) (*Knuu, error) {
 		if err := setupProxy(ctx, k); err != nil {
 			return nil, err
 		}
-	}
-
-	if err := k.handleTimeout(ctx); err != nil {
-		return nil, err
 	}
 
 	return k, nil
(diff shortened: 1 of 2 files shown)


pkg/knuu/knuu.go
```diff
@@ -49,8 +49,6 @@ type Options struct {
 }
 
 func New(ctx context.Context, opts Options) (*Knuu, error) {
-	opts.TestScope = k8s.SanitizeName(opts.TestScope)
```
mojtaba-esk, line 52: This was supposed to be removed in an already merged PR, but it is still there an causing panics. So, it is removed even though it is not relevant to this PR.

smuu, line 52: Removing this will break the app tests/simulations again.

The app repo still uses the deprecated version to initialize knuu. This results in a mismatch of the scope and the namespace.

The namespace is sanitized but not the scope.

smuu (changes requested):

pkg/knuu/knuu.go
smuu: If this causes panic because `SanitizeName` fails if `opts.TestScope` is an empty string, we could add a check for that case.

mojtaba-esk: The decision was to set a default scope when it is empty and it is checked in the `setDefaults` func. That's why we have it removed from the beginning of the `New` func.

mojtaba-esk: Also the scope is sanitized in the `setDefault` func here: https://github.com/celestiaorg/knuu/blob/07440a77e606ab29976d42c050f2b47c08dd87d7/pkg/knuu/knuu.go#L204

The app tests should not fail. If they do, can you please point to the failing tests case, so we can have a look and see why it is failing?

smuu: I can't reproduce the issue anymore. So should be fine if we remove this line 👍

smuu (approved):

PR #503, opened 2024-07-15
chore!: removing the loadEnvVars func

### LINKED ISSUE #457: Consider removing or moving the loadEnvVars from New to somewhere else
loading env var is not the main task of knuu as with the recent refactors, it is configurable via its `New` function.
So this issue suggests that to remove it from `knuu.New` and allow users to load it when they need for their purpose and then configure the `knuu` the way that they need.

mojtaba-esk:
Closes #457 

## Summary by CodeRabbit

- **New Features**
	- Enhanced documentation regarding the handling of environment variables for the `knuu` tool.
  
- **Bug Fixes**
	- Streamlined error handling with clearer naming conventions for scope-related errors.
  
- **Documentation**
	- Clarified configuration options for environment variables in the README.

- **Chores**
	- Updated module dependencies and Go version to improve performance and compatibility.


MSevey:
We have documentation that tells users to use environment variables. https://github.com/celestiaorg/knuu?tab=readme-ov-file#environment-variables

If we have documentation that references env vars, imo we should: 
- also include documentation on how to access those env vars if we are removing this functionality internal to knuu
- have testing of those env vars (which should have broken with this PR if they existed)

I don't have a strong opinion on whether or not this functionality should be removed with this PR or stay, but either way the following should be true. 

1. We have testing that validates the use case of the environment variables we have in our documentation.
2. Our documentation clearly reflects who/what is responsible for loading the env vars.

MSevey (changes requested):
see comment

mojtaba-esk:
> We have documentation that tells users to use environment variables. https://github.com/celestiaorg/knuu?tab=readme-ov-file#environment-variables

Thanks for pointing this out. This PR removes the function that loads `.env` var by default which is not a Job for knuu, but for user to decide how they wanna configure their ENV vars.

We need to update that document and there is an issue for it to track it, right now the only env var our of the 3 listed in the doc is available and it is `LOG_LEVEL`

smuu (changes requested):
Could we add some documentation for that to this PR before merging it?

mojtaba-esk:
ok docs updated according to the changes @MSevey @smuu

MSevey (dismissed):
Can you update this PR to target main?

mojtaba-esk:
> Can you update this PR to target main?

sure, let's first approve & merge this PR: https://github.com/celestiaorg/knuu/pull/502 so there will be less conflict resolving work afterwards.

smuu (dismissed):

smuu (dismissed):

smuu (approved):

PR #508, opened 2024-07-16
chore!: remove Executor from knuu

### LINKED ISSUE #467: deprecate Executor

mojtaba-esk:
Closes #467


e2e/executor.go
```diff
@@ -0,0 +1,53 @@
...
+	sleepCommand         = "sleep"
+	infinityArg          = "infinity"
+)
+
+var (
+	executorMemoryLimit = resource.MustParse("100Mi")
+	executorCpuLimit    = resource.MustParse("100m")
+)
+
+func NewExecutor(ctx context.Context, executorName string) (*knuu.Instance, error) {
```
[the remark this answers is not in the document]
mojtaba-esk: No this is not created for compatibility. It is moved to the e2e pkg to be served as a utility function to avoid redundant codes in our tests.

smuu (approved):
I changed the title as it's a breaking change and it's removing the Executor and not deprecating it.

smuu (approved):

pkg/instance/type.go
```diff
@@ -7,7 +7,6 @@ type InstanceType int
 const (
 	UnknownInstance InstanceType = iota
 	BasicInstance
-	ExecutorInstance
```
smuu, line 10: This gave us the possibility to disable networking between Instances without disabling networking from and to executor instances. We would lose this ability.

The question is if we need the scenario where we disable the networking of a node for a certain time to simulate network connection loss.

smuu (changes requested):

pkg/instance/type.go
mojtaba-esk: I see. Since we are removing executor from the instance package, we might want to define another type of instance which resists the network disabling tasks. Something like `AlwaysEnabledNetworkInstance` not very nice name :D but you get the point.

mojtaba-esk: If you agree with this idea, let's track it in a separate issue

smuu: That makes sense 👍 We need to document that in a very clear way to the dev.

mojtaba-esk: an issue is created for it #522

smuu (approved):

PR #515, opened 2024-07-19
chore!: set image and set command require a running instance to stop

### LINKED ISSUE #484: SetImage and SetCommand should require a instance.Stop()

mojtaba-esk:
Closes #484 

## Summary by CodeRabbit

- **New Features**
	- Enhanced error messages for command settings to clarify state restrictions.
	- Streamlined logic for setting images and commands, improving state management.
- **Deprecations**
	- Removed the `SetImageInstant` method and marked the `SetCommand` method as deprecated.
- **Bug Fixes**
	- Improved control flow in the commit operation to ensure logical sequence in policy rule addition.
- **Chores**
	- Removed the `setImageWithGracePeriod` method to simplify instance lifecycle management.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -230,19 +230,14 @@ func (i *Instance) SetInstanceType(instanceType InstanceType) {
 }
 
 // SetImage sets the image of the instance.
-// When calling in state 'Started', make sure to call AddVolume() before.
-// It is only allowed in the 'None' and 'Started' states.
+// It is not allowed in the 'Committed' and 'Started' states.
 func (i *Instance) SetImage(ctx context.Context, image string) error {
-	if !i.IsInState(StateNone, StateStarted) {
+	if i.IsInState(StateCommitted, StateStarted) {
 		return ErrSettingImageNotAllowed.WithParams(i.state.String())
 	}
 
 	if i.isSidecar {
 		return ErrSettingImageNotAllowedForSidecarsStarted
-	}
-
-	if i.state == StateStarted {
-		return i.setImageWithGracePeriod(ctx, image, nil)
 	}
 
 	// Use the builder to build a new image
@@ -282,25 +277,10 @@ func (i *Instance) SetGitRepo(ctx context.Context, gitContext builder.GitContext
 	return i.builderFactory.BuildImageFromGitRepo(ctx, gitContext, imageName)
 }
 
-// SetImageInstant sets the image of the instance without a grace period.
-// Instant means that the pod is replaced without a grace period of 1 second.
-// It is only allowed in the 'Running' state.
-func (i *Instance) SetImageInstant(ctx context.Context, image string) error {
-	if !i.IsState(StateStarted) {
-		return ErrSettingImageNotAllowedForSidecarsStarted.WithParams(i.state.String())
-	}
-
-	if i.isSidecar {
-		return ErrSettingImageNotAllowedForSidecars
-	}
-
-	return i.setImageWithGracePeriod(ctx, image, nil)
-}
-
 // SetCommand sets the command to run in the instance
 // This function can only be called when the instance is in state 'Preparing' or 'Committed'
 func (i *Instance) SetCommand(command ...string) error {
-	if !i.IsInState(StatePreparing, StateCommitted) {
+	if i.IsInState(StateStarted, StateCommitted) {
 		return ErrSettingCommand.WithParams(i.state.String())
 	}
 	i.command = command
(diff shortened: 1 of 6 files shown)


pkg/instance/instance.go
```diff
@@ -230,21 +230,16 @@ func (i *Instance) SetInstanceType(instanceType InstanceType) {
 }
 
 // SetImage sets the image of the instance.
-// When calling in state 'Started', make sure to call AddVolume() before.
-// It is only allowed in the 'None' and 'Started' states.
+// It is not allowed in the 'Committed' and 'Started' states.
 func (i *Instance) SetImage(ctx context.Context, image string) error {
-	if !i.IsInState(StateNone, StateStarted) {
```
smuu, line 236: There are two use cases for `SetImage`
1. To get from `StateNone` to `StatePreparing` after a new instance was created
2. To restart an Instance with a new image after the Instance already started.

smuu: Removing the `StateNone` here means we can't get from `StateNone` to `StatePreparing` by setting an existing image.

pkg/instance/instance.go
```diff
@@ -230,21 +230,16 @@ func (i *Instance) SetInstanceType(instanceType InstanceType) {
...
 	}
 
 	if i.isSidecar {
 		return ErrSettingImageNotAllowedForSidecarsStarted
 	}
 
-	if i.state == StateStarted {
-		return i.setImageWithGracePeriod(ctx, image, nil)
-	}
-
```
smuu, line 247: Removing this means we can't use `SetImage` anymore to replace an already running Instance with a new image.

smuu: As using the function `SetImage` in two different cases might be confusing, we might want to consider creating another function like `UpgradeImage`, which allows the use case of changing the image.

pkg/instance/instance.go
```diff
@@ -282,25 +277,10 @@ func (i *Instance) SetGitRepo(ctx context.Context, gitContext builder.GitContext
...
-	return i.setImageWithGracePeriod(ctx, image, nil)
-}
-
 // SetCommand sets the command to run in the instance
 // This function can only be called when the instance is in state 'Preparing' or 'Committed'
 func (i *Instance) SetCommand(command ...string) error {
-	if !i.IsInState(StatePreparing, StateCommitted) {
+	if i.IsInState(StateStarted, StateCommitted) {
```
smuu, line 283: Once the Instance is started, `SetCommand` will not have any effect.
This function is used to set the command of which the container is started, which we need to be able to set in `StatePreparing`.

pkg/instance/instance.go
```diff
@@ -282,25 +277,10 @@ func (i *Instance) SetGitRepo(ctx context.Context, gitContext builder.GitContext
...
-	if !i.IsState(StateStarted) {
-		return ErrSettingImageNotAllowedForSidecarsStarted.WithParams(i.state.String())
-	}
-
-	if i.isSidecar {
-		return ErrSettingImageNotAllowedForSidecars
-	}
-
-	return i.setImageWithGracePeriod(ctx, image, nil)
-}
```
smuu, line 298: When calling `SetImage` in `StateStarted`, it uses a grace period by default. There is a use case where you want to set this grace period to 0. This is why this function was introduced.

smuu: If we would introduce a function like `UpgradeImage`, we could let the user pass an optional parameter to allow to set the grace period of upgrading the image.

smuu:
@mojtaba-esk Do you remember why we wanted to change this. I can't remember the reason.

mojtaba-esk:
> @mojtaba-esk Do you remember why we wanted to change this. I can't remember the reason.

Not really, but there is [an issue for it here ](https://github.com/celestiaorg/knuu/issues/484) unfortunately there is no more context about it there. We probably wanted to do it to keep things consistent and tidy.

smuu:
> > @mojtaba-esk Do you remember why we wanted to change this. I can't remember the reason.
> 
> Not really, but there is [an issue for it here ](https://github.com/celestiaorg/knuu/issues/484) unfortunately there is no more context about it there. We probably wanted to do it to keep things consistent and tidy.

I thought about it again, and I now believe that we have the grouping; this is not needed anymore.

mojtaba-esk:
> > > @mojtaba-esk Do you remember why we wanted to change this. I can't remember the reason.
> > 
> > 
> > Not really, but there is [an issue for it here ](https://github.com/celestiaorg/knuu/issues/484) unfortunately there is no more context about it there. We probably wanted to do it to keep things consistent and tidy.
> 
> I thought about it again, and I now believe that we have the grouping; this is not needed anymore.

So then let's close this issue.

PR #530, opened 2024-07-25
chore!: use unique static names for instances

### LINKED ISSUE #526: Names must be static (scope & instance name) -> let user define it

### LINKED ISSUE #527: remove random suffix from the k8s name

### LINKED ISSUE #528: knuu object keeps map of all instances -> every knuu run user cannot create two instances with the same name

mojtaba-esk:
Closes #526
Closes #527 
Closes #528 

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for instance management with new error variables for duplicate names.
	- Added thread-safe methods for managing instance names within the system dependencies.
	- Implemented unique name checks for various methods in the Traefik service.

- **Bug Fixes**
	- Improved naming conventions across multiple test cases to enhance clarity and prevent conflicts.

- **Refactor**
	- Standardized naming conventions within the Instance struct and its methods.
	- Streamlined the Preloader struct to use a consistent naming scheme.
	- Improved the Traefik struct to manage unique names effectively.

- **Documentation**
	- Updated error messages to provide clearer feedback on specific error scenarios.

### DIFF
--- a/pkg/instance/sidecars.go
+++ b/pkg/instance/sidecars.go
@@ -7,10 +7,12 @@ import (
 )
 
 type SidecarManager interface {
-	Initialize(ctx context.Context, sysDeps *system.SystemDependencies) error
+	// namePrefix is the name of the instance that is the parent of the sidecar
+	// it is used to set the name of the sidecar to avoid name collisions
+	Initialize(ctx context.Context, namePrefix string, sysDeps *system.SystemDependencies) error
 	Instance() *Instance
 	PreStart(ctx context.Context) error
-	Clone() (SidecarManager, error)
+	Clone(namePrefix string) (SidecarManager, error)
 }
 
 type sidecars struct {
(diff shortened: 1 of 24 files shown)


smuu (changes requested):
Thank you for already working on this.
I would like to have this for `v0.16.0` as I want to properly test this and I don't want to rush it to get it into `v0.15.0`.

mojtaba-esk:
> Thank you for already working on this. I would like to have this for `v0.16.0` as I want to properly test this and I don't want to rush it to get it into `v0.15.0`.

Ok, so I'll move it down in the prio list

pkg/instance/sidecars.go
```diff
@@ -92,12 +94,16 @@ func (s *sidecars) setStateForSidecars(state InstanceState) {
 		})
 }
 
-func (s *sidecars) cloneWithSuffix(suffix string) *sidecars {
+func (s *sidecars) clone(namePrefix string) (*sidecars, error) {
```
MSevey, line 97: Why the change from suffix to prefix?
Doesn't suffix make it easier to see everything that is related? printed out alphabetically?

MSevey (dismissed):

pkg/instance/sidecars.go
mojtaba-esk: When clone an instance, we need to clone also the sidecars and to avoid name collision, we use the instance name as the prefix for its sidecar's names.

pkg/system/dependencies.go
```diff
@@ -17,4 +19,18 @@ type SystemDependencies struct {
 	Proxy        *traefik.Traefik
 	Scope        string
 	StartTime    string
+	instancesMap sync.Map
```
smuu, line 22: What is the reason we use a `Map` here as we don't use the value of the `Map`?

smuu (dismissed):

pkg/system/dependencies.go
mojtaba-esk: it is for the quick search in its keys. bc if we use a slice, then we have to iterate over all of it to get the answer, but a Map uses a hash-table which is very quick in finding the keys.
Here we also use a `sync.Map` which is multi thread safe compare to the standard go map.

PR #533, opened 2024-07-26
chore: group instance methods

### LINKED ISSUE #519: Consider grouping of instance functions.
Groups to introduce: `networkShaping` and `observability`

To call functions of such a group, the dev would call: `instance.<groupName>.<functionName>`

This could improve the understanding of the usage of different functions.

networkShaping
- EnableBitTwister/EnableNetworkShaping
- DisableBitTwister/DisableNetworkShaping
- SetLatencyAndJitter
- SetPacketLoss
- SetBandwidthLimit

observability:
- SetJaegerEndpoint
- SetJaegerExporter
- SetOtelCollectorVersion
- SetOtelEndpoint
- SetOtlpExporter
- SetPrometheusEndpoint
- SetPrometheusExporter
- SetPrometheusRemoteWriteExporter

Please let me know what you think @mojtaba-esk and @tty47

mojtaba-esk:
Closes #519 

## Summary by CodeRabbit

- **New Features**
  - Introduced a structured approach to manage container instance lifecycle within a Kubernetes environment, enhancing modularity and encapsulation.
  - Added methods to set container image, commands, and environment variables, streamlining instance configuration.
  - Implemented a comprehensive execution management system to control starting, stopping, and destroying instances.
  - Enhanced security management capabilities with methods to configure privileged mode and add container capabilities.

- **Bug Fixes**
  - Improved error handling and state management during image builds and command executions.

- **Refactor**
  - Restructured the `Instance` type to include new modular components, promoting better organization and clearer responsibilities.
  - Updated method calls across tests for better clarity and consistency, transitioning to the new modular design.

- **Documentation**
  - Updated comments and documentation to reflect the new modular structure and usage of methods.


pkg/instance/execution.go
smuu, line 1: The old function `SetImage()` (and `SetImageInstant()`) was used for two different use cases, one for initializing an Instance and for replacing a running instance with a new image. This PR removed this feature.
It seems there is no test for that, we should add one.

This is how the app team uses to upgrade the image for an instance: https://github.com/celestiaorg/celestia-app/blob/4f4d0f7ff1a43b62b232726e52d1793616423df7/test/e2e/minor_version_compatibility.go#L87

smuu (changes requested):
I will continue the review later.

pkg/instance/resources.go
```diff
@@ -0,0 +1,157 @@
...
+
+// AddPolicyRule adds a policy rule to the instance
+// This function can only be called in the states 'Preparing' and 'Committed'
+func (r *resources) AddPolicyRule(rule rbacv1.PolicyRule) error {
+	if !r.instance.IsInState(StatePreparing, StateCommitted) {
+		return ErrAddingPolicyRuleNotAllowed.WithParams(r.instance.state.String())
+	}
+	r.policyRules = append(r.policyRules, rule)
+	return nil
+}
```
smuu, line 31: Is there a reason this in in `resources` and not in `security`?

smuu (changes requested):
This makes so much sense! Thank you for this PR.
Only a few small things.

pkg/instance/execution.go
mojtaba-esk: There was a discussion on not letting user to change image of a running instance. which means the user has to `Stop` the instance and then set a new image and then `Start` again. This is towards that. Do we want to keep that functionality?

mojtaba-esk: I saw the code they have to update the following function:

```go
func (n *Node) Upgrade(version string) error
```
to stop the instance, update the image and start again. and it should work as before.

pkg/instance/resources.go
mojtaba-esk: There is no reason for the group names of the functions. I just put them in the groups that I thought it might make sense. Since you guys have more experience with `k8s`, please feel free to move the funcs in the groups that are more relevant or just comment on them and I will move them.

pkg/instance/execution.go
smuu: Now I remember that discussion 👍 We can work in a seperate PR on that then.

`SetImage()` is now part of `Build()`. Not sure if this makes sense when we stop and upgrade the image inside that group then.
Would it make sense to have `SetImage()` in `Build()` and only allow this when preparing the Instance and then another function in another group to upgrade the image after it was stopped.

pkg/instance/resources.go
smuu: Okay, I think this function makes more sense in `security`.

pkg/instance/execution.go
```diff
@@ -0,0 +1,437 @@
...
+	return nil
+}
+
+func (e *execution) UpgradeImage(ctx context.Context, image string) error {
+	if !e.instance.IsInState(StateStarted) {
+		return ErrUpgradingImageNotAllowed.WithParams(e.instance.state.String())
+	}
+
+	return e.instance.build.setImageWithGracePeriod(ctx, image, nil)
+}
```
smuu, line 275: Allow the dev to specify the grace period

smuu (changes requested):
Only one small thing, otherwise LGTM

pkg/instance/execution.go
mojtaba-esk: done

pkg/instance/security.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (s *security) clone() *security {
+	capabilitiesAddCopy := make([]string, len(s.capabilitiesAdd))
+	copy(capabilitiesAddCopy, s.capabilitiesAdd)
+
+	return &security{
+		instance:        nil,
+		privileged:      s.privileged,
+		capabilitiesAdd: capabilitiesAddCopy,
+	}
```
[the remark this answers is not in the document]
mojtaba-esk: at this stage we do not have a reference to the new instance as it does not exist yet, so we set it to `nil` and it is user's responsibility to set it once it is created.

smuu (dismissed):
Nice job 🪨 We might have missed something but we can fix that once we test the `rc` in the app repo.

smuu (approved):

Issue #523, opened 2024-07-24
Fix errors in `e2e/system`


tty47:
Similar to this one:

https://github.com/celestiaorg/knuu/actions/runs/10076571857/job/27857216524#step:5:129

I was able to reproduce it locally finding some more errors

cc: @mojtaba-esk  I'll keep you posted about it

PR #535, opened 2024-08-06
fix(e2e-523): inital commit - fixing  e2e/system tests  - WIP

tty47:
closes: https://github.com/celestiaorg/knuu/issues/523

## Summary by CodeRabbit

- **New Features**
  - Enhanced error handling in testing functions with a new retry mechanism, improving test reliability against transient errors.
  - Introduced a utility function for retry operations, facilitating multiple attempts for potentially failing actions.
  - Implemented synchronization mechanisms for concurrent access to bucket creation within the Minio service.
  - Added new configuration options for container image pull policies, enhancing deployment flexibility.
  - Increased timeout for specific GitHub Actions workflows to accommodate longer-running test processes.
  - Removed the `TestNoVolumesNoFiles` function and expanded the `TestOneVolumeOneFile` function to handle two files, enhancing the test suite's focus.

- **Bug Fixes**
  - Improved robustness of test scenarios by addressing potential hanging issues through timeout implementations.

- **Refactor**
  - Streamlined test logic and renamed functions for clearer representation of test cases.
  - Removed outdated test functions to focus on relevant scenarios involving volumes and files.
  - Enhanced logging and error messages within tests for better debugging and traceability.


mojtaba-esk:
Thanks for the great work. Now all tests are passing. Would you please resolve conflicts so we will be able to merge it?

e2e/system/build_from_git_test.go
```diff
@@ -12,10 +14,14 @@ func (s *Suite) TestBuildFromGit() {
 	s.T().Parallel()
 
 	// Setup
-	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
```
mojtaba-esk, line 17: Why 10 mins is used? if we use `context.Background()` the test timeout is used which is by default 10 mins or whatever the user sets

e2e/system/build_from_git_test.go
```diff
@@ -12,10 +14,14 @@ func (s *Suite) TestBuildFromGit() {
...
+	defer cancel()
 
+	s.T().Log("Creating new instance")
 	target, err := s.Knuu.NewInstance(namePrefix)
-	s.Require().NoError(err)
+	if err != nil {
+		s.T().Fatalf("Error creating new instance: %v", err)
+	}
```
mojtaba-esk, line 24: `s.Require().NoError(err)` does the same thing in one line. We can also set a custom message for it:

```go
s.Require().NoError(err, "some message")
```

e2e/system/build_from_git_test.go
```diff
@@ -26,66 +32,130 @@ func (s *Suite) TestBuildFromGit() {
 		Username: "",
 		Password: "",
 	})
-	s.Require().NoError(err)
+	if err != nil {
+		s.T().Fatalf("Error setting git repo: %v", err)
+	}
```
mojtaba-esk, line 37: same here

e2e/system/build_from_git_test.go
```diff
@@ -26,66 +32,130 @@ func (s *Suite) TestBuildFromGit() {
...
 		if err := target.Destroy(ctx); err != nil {
 			s.T().Logf("Error cleaning up knuu: %v", err)
 		}
 	})
 
-	s.Require().NoError(target.Commit())
+	s.T().Log("Committing changes")
+	if err := target.Commit(); err != nil {
+		s.T().Fatalf("Error committing changes: %v", err)
+	}
```
mojtaba-esk, line 51: also here and the rest of the code.

e2e/system/build_from_git_test.go
```diff
@@ -26,66 +32,130 @@ func (s *Suite) TestBuildFromGit() {
...
-		Repo:     "https://github.com/celestiaorg/knuu.git",
-		Branch:   "test/build-from-git", // This branch has a Dockerfile and is protected as to not be deleted
-		Username: "",
-		Password: "",
-	})
-	s.Require().NoError(err)
-
-	s.Require().NoError(target.SetCommand("sleep", "infinity"))
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
```
mojtaba-esk, line 85: same comment about ctx

e2e/system/build_from_git_test.go
```diff
@@ -26,66 +32,130 @@ func (s *Suite) TestBuildFromGit() {
...
+		return target.Commit()
+	}, maxRetries)
+	if err != nil {
+		s.T().Fatalf("Error committing changes: %v", err)
+	}
 
 	s.T().Cleanup(func() {
+		s.T().Log("Cleaning up instance")
 		if err := target.Destroy(ctx); err != nil {
 			s.T().Logf("Error cleaning up knuu: %v", err)
```
mojtaba-esk, line 133: there was a typo in the message from before. it should be something like this:
```go
s.T().Logf("error destroying instance: %v", err)
```

e2e/system/build_from_git_test.go
```diff
@@ -26,66 +32,130 @@ func (s *Suite) TestBuildFromGit() {
...
+		data, err = target.GetFileBytes(ctx, "/home/hello.txt")
+		return err
+	}, maxRetries)
+	if err != nil {
+		s.T().Fatalf("Error getting file bytes: %v", err)
+	}
+
+	if !bytes.Equal([]byte("Hello, world!"), data) {
+		s.T().Fatalf("File bytes do not match. Expected 'Hello, world!', got '%s'", string(data))
+	}
```
mojtaba-esk, line 158: do these lines create a different outcome from the following line?

```go
s.Assert().Equal([]byte("Hello, world!"), data, "File bytes do not match")
```

e2e/system/suite_setup_test.go
```diff
@@ -94,3 +97,19 @@ func (s *Suite) createNginxInstanceWithVolume(ctx context.Context, name string)
...
+			return err
+		}
+
+		conn, err := net.DialTimeout("tcp", net.JoinHostPort(ip, fmt.Sprintf("%d", nginxPort)), 5*time.Second)
+		if err != nil {
+			return err
+		}
+		conn.Close()
+		return nil
+	}, 10)
```
mojtaba-esk, line 114: lets extract also that `10` constant

e2e/system/utils.go
```diff
@@ -0,0 +1,18 @@
...
+	var err error
+	for i := 0; i < maxRetries; i++ {
+		err = operation()
+		if err == nil {
+			return nil
+		}
+		time.Sleep(time.Second * time.Duration(i+1))
+	}
+	return fmt.Errorf("operation failed after %d retries: %w", maxRetries, err)
+}
```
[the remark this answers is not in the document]
mojtaba-esk: A valid suggestion. Since we are using it in test context, we can use the `s.T().Log...` thing

mojtaba-esk (changes requested):
Thanks for the great work. Adding `retryOperation` is a smart choice. 
Left some comments and suggestions.

e2e/system/build_from_git_test.go
mojtaba-esk: I suggest to use `context.Background()` and when running the test use `-timeout=20m` flag to increase the test timeout. that should work.

mojtaba-esk: the `20m` is an example, use any value that you think is safe to be running on most of our platforms

smuu (dismissed):

mojtaba-esk (approved):

e2e/system/build_from_git_test.go
```diff
@@ -1,32 +1,41 @@
...
 
 	// Setup
 	ctx := context.Background()
 
+	s.T().Log("Creating new instance")
 	target, err := s.Knuu.NewInstance(namePrefix)
-	s.Require().NoError(err)
+	if err != nil {
+		s.Require().NoError(err, "Error creating new instance")
+	}
```
MSevey, line 27: the if statement is redundant. 

```suggestion
	s.Require().NoError(err, "Error creating new instance")
```

e2e/system/build_from_git_test.go
```diff
@@ -41,51 +50,79 @@ func (s *Suite) TestBuildFromGit() {
...
 	// The file is created by the dockerfile in the repo,
 	// so to make sure it is built correctly, we check the file
 	data, err := target.Storage().GetFileBytes(ctx, "/test.txt")
 	s.Require().NoError(err)
 
 	data = []byte(strings.TrimSpace(string(data)))
-	s.Assert().Equal([]byte("Hello, World!"), data, "File bytes do not match")
+	if !bytes.Equal([]byte("Hello, World!"), data) {
+		s.Require().NoError(err, "File bytes do not match. Expected 'Hello, World!', got '%s'", string(data))
+	}
```
MSevey, line 64: This should be reverted. The Assert statement was handling the byte comparison. In this set up, we are actually completely ignoring the byte mismatch because if you get to this point, the `err` is nil because of the check above. So this wouldn't stop the test.

e2e/system/build_from_git_test.go
```diff
@@ -41,51 +50,79 @@ func (s *Suite) TestBuildFromGit() {
...
+
+	s.T().Log("Test completed successfully")
```
MSevey, line 66: remove, unnecessary.

e2e/system/build_from_git_test.go
```diff
@@ -41,51 +50,79 @@ func (s *Suite) TestBuildFromGit() {
...
-		Username: "",
-		Password: "",
-	})
-	s.Require().NoError(err)
+
+	s.T().Log("Creating new instance")
+	target, err := s.Knuu.NewInstance(namePrefix)
+	if err != nil {
+		s.Require().NoError(err, "Error creating new instance")
+	}
```
MSevey, line 83: same, you don't need if statements when using require. 

```suggestion
	s.Require().NoError(err, "Error creating new instance")
```

e2e/system/file_test.go
```diff
@@ -13,44 +13,79 @@ import (
...
 
+	// Setup
 	ctx := context.Background()
+
+	s.T().Log("Creating executor instance")
 	executor, err := s.Executor.NewInstance(ctx, namePrefix+"-executor")
-	s.Require().NoError(err)
+	if err != nil {
+		s.Require().NoError(err, "Error creating executor instance")
+	}
```
MSevey, line 29: same, don't use if statement

e2e/system/files_to_volumes_cm_test.go
```diff
@@ -11,43 +11,6 @@ import (
 	"github.com/celestiaorg/knuu/pkg/instance"
 )
 
-// TestOneVolumeNoFiles tests the scenario where we have one volume and no files.
-// the initContainer command that it generates looks like:
-// no initContainer command, as there is no volumes, nor files.
-func (s *Suite) TestNoVolumesNoFiles() {
```
MSevey, line 17: Why did we delete this entire test?

pkg/builder/kaniko/kaniko.go
```diff
@@ -31,7 +31,6 @@ const (
 
 type Kaniko struct {
 	system.SystemDependencies
-	ContentName string // Name of the content pushed to Minio
```
MSevey, line 34: Why did we remove this?

pkg/minio/minio.go
```diff
@@ -322,6 +324,18 @@ func (m *Minio) createBucketIfNotExists(ctx context.Context, bucketName string)
...
+	}
+
+	mu, ok := m.muMap[bucketName]
+	if !ok {
+		mu = &sync.Mutex{}
+		m.muMap[bucketName] = mu
+	}
+	mu.Lock()
+	defer mu.Unlock()
+
```
MSevey, line 338: We should have some map clean up code on the error path. 

you need to add a variable name to the return error like this. 
```go
func (m *Minio) createBucketIfNotExists(ctx context.Context, bucketName string) (err error) {
```
then you could add a defer statement
```suggestion
defer func() {
   if err != nil {
      delete(mu.Map, bucketname)
   }
} ()
```

This would protect us from accessing the map and expecting the bucket to exist when it doesn't due to an error.

MSevey (changes requested):
Good use of the retry func

e2e/system/files_to_volumes_cm_test.go
mojtaba-esk: @tty47 I have the same question as well. Is it not needed anymore?

pkg/builder/kaniko/kaniko.go
mojtaba-esk: concurrency issues. it is moved to somewhere else, where it is used.

pkg/minio/minio.go
mojtaba-esk: The deletion is not necessary as it is unlocked on return even if it is due to an error.

```go
defer mu.Unlock()
```

Currently there are max 2 buckets created in a complete knuu run. But clean up is usually a good thing.

Why do we need to have var name `err` in the signature?

MSevey: The issues wasn't that use of the lock, but protecting against the error case of trying to use a bucket in the map what failed to be created.

MSevey: In this instance, adding the var name for the return error ensures that we are checking that error during the defer func. Also where we would want to define the defer func, an error hasn't been defined yet, so we need to define one anyways.

e2e/system/build_from_git_test.go
MSevey: not resolved

e2e/system/build_from_git_test.go
MSevey: not resolved

e2e/system/build_from_git_test.go
MSevey: not resolved

e2e/system/build_from_git_test.go
MSevey: not resolved

e2e/system/file_test.go
MSevey: not resolved

mojtaba-esk: done

pkg/minio/minio.go
mojtaba-esk: got it, in the proposed defer func, we wanted to cleanup the Map and then each call causes it to create a new entry for the same key which is not very optimized. So we need to decide on that. I prefer to keep it without defer for sake of optimization as that function is called for every file push.

e2e/system/build_from_git_test.go
mojtaba-esk: This is outdated, it is already fixed.

e2e/system/files_to_volumes_cm_test.go
```diff
@@ -240,28 +205,34 @@
...
+				err := i.Storage().AddFile(resourcesFileCMToFolder+"/test_1", nginxHTMLPath+"/index.html", "0:0")
+				if err != nil {
+					return fmt.Errorf("adding file to '%v': %w", i.Name(), err)
+				}
+				err = i.Storage().AddFile(resourcesFileCMToFolder+"/test_2", nginxHTMLPath+"/index-2.html", "0:0")
+				if err != nil {
+					return fmt.Errorf("adding file to '%v': %w", i.Name(), err)
+				}
+				return nil
+			}, maxRetries)
```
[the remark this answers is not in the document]
mojtaba-esk: it is there, you did not see it. The tests are passing.

MSevey (approved):

PR #549, opened 2024-08-23
chore: tiny optimizations on netshaper tests

mojtaba-esk:
This PR proposes a set of small optimizations and cleanup procedure to run the netshaper tests smoother and faster

Please note that the system tests are filing due to a series of issues that are resolved in #535  

## Summary by CodeRabbit

- **New Features**
  - Enhanced test management with systematic instance naming to improve clarity and reduce name collisions.
  - Introduced mechanisms for thread-safe test cleanup, optimizing resource management and execution flow.
  
- **Bug Fixes**
  - Improved synchronization during test cleanup to ensure resources are released efficiently after all tests complete.

### DIFF
--- a/e2e/netshaper/suite_setup_test.go
+++ b/e2e/netshaper/suite_setup_test.go
@@ -2,9 +2,10 @@ package netshaper
 
 import (
 	"context"
+	"sync"
+	"sync/atomic"
 	"testing"
 
-	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/suite"
 
 	"github.com/celestiaorg/knuu/pkg/knuu"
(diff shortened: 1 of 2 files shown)


e2e/netshaper/suite_setup_test.go
```diff
@@ -13,6 +14,14 @@ import (
...
 	Knuu *knuu.Knuu
+
+	cleanupMu     sync.Mutex
+	totalTests    atomic.Int32
+	finishedTests int32
+}
+
+func TestRunSuite(t *testing.T) {
+	suite.Run(t, new(Suite))
 }
```
MSevey, line 25: Follow Up: Can this be refactored to just be in the e2e package? We should be able to use the same struct across the tests as opposed to redefining it.

MSevey (approved):

e2e/netshaper/suite_setup_test.go
mojtaba-esk: Good idea, yeah let's track it in a followup issue.

PR #563, opened 2024-09-13
chore!: adjust stopped status actions for the update image and others

### LINKED ISSUE #561: feat: set the args when we replace the image
## Problem:
When replacing an image or switching the binary, there seems to be no straightforward way to set additional arguments, such as the `--upgrade-v2-height` flag, when the new image is run.

## Current Behavior:
Setting arguments using `SetArgs` before upgrading the image should work; however, there's a status check that prevents setting args unless the instance is in either the `Preparing` or `Committed` state.

## Proposal solution:
User can Stop the instance, Set the args and then Upgrade the image. To allow that to happen, we need to allow SetArgs to accept `StateStopped` and not only SetArgs but all funcs that user could call during that period.

mojtaba-esk:
Closes #561 

Note: This PR removes the `UpgradeImage` functions, instead, the user has to stop their instance, configure whatever they want and start it again.

## Release Notes

- **New Features**
  - Introduced flexibility in state management, allowing operations like setting images, adding ports, and adjusting resources even when instances are in a 'Stopped' state.
  - Enhanced error handling with clearer feedback for state restrictions when adding hosts to a proxy.

- **Bug Fixes**
  - Improved state checks for various methods, ensuring proper functionality across different instance states.

- **Documentation**
  - Updated comments and documentation to reflect changes in state conditions for various functionalities.


MSevey (changes requested):
A breaking change should break tests. If not then that is a clear gap in our testing, so we should be adding a test for the new use case.

mojtaba-esk:
> A breaking change should break tests. If not then that is a clear gap in our testing, so we should be adding a test for the new use case.

There is no  dedicated E2E test for the upgrade image that's why it does not brake any. There is one use case in the app e2e test which we have to update the sequence. i.e. Stop the instance first, upgrade the image, then Start the instance again.

smuu (dismissed):

smuu (approved):

MSevey (approved):
Approved with https://github.com/celestiaorg/knuu/issues/565 as a follow up

PR #566, opened 2024-09-25
fix: update configmap if exists

mojtaba-esk:
While attempting to upgrade app e2e tests to knuu 0.16.0-rc.0, found out that some of the upgrade tests fail due to existing configmap and the main reason is because we use static names and it is expected to update a resource if it already exists. This PR proposes a change to the configmaps if they already exist which helps the tests pass when an upgrade is happening.

## Summary by CodeRabbit

- **New Features**
	- Enhanced functionality for managing Kubernetes ConfigMaps with methods to create or update them seamlessly.
	- Introduced validation for ConfigMap parameters to ensure data integrity.
- **Bug Fixes**
	- Improved error handling for ConfigMap updates, providing clearer error messages for failures.

### DIFF
--- a/pkg/instance/storage.go
+++ b/pkg/instance/storage.go
@@ -2,7 +2,6 @@ package instance
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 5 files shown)


smuu (commented):
Please remember that when changing a `ConfigMap`, the `Pod` the `ConfigMap` is mounted to, is not restarted automatically.
So we should consider to restart the `Pod`(`Instance`) when the dev only changes a `ConfigMap`.

mojtaba-esk:
> Please remember that when changing a `ConfigMap`, the `Pod` the `ConfigMap` is mounted to, is not restarted automatically. So we should consider to restart the `Pod`(`Instance`) when the dev only changes a `ConfigMap`.

Great point. Thanks for raising it. 
[I double checked it](https://github.com/celestiaorg/knuu/blob/mojtaba/fix-update-configmap-if-exist/pkg/instance/execution.go#L110-L111) and it is called only when the instance is in the `committed` state, so even if user wants to change it somehow, they need to stop the instance do changes then commit it and then start so it will be called. 
We have to bare that in mind for future development.

MSevey (approved):

PR #570, opened 2024-10-08
chore: add a check for Instance state when cloning

### LINKED ISSUE #531: feat: `CloneWithSuffix` does not check for Instance state
`Clone` checks for state Instance is in but `CloneWithSuffix` does not.

https://github.com/celestiaorg/knuu/blob/f21844326cbab7ac02332612a2aafddadd9ea0dd/pkg/instance/instance.go#L1074

https://github.com/celestiaorg/knuu/blob/f21844326cbab7ac02332612a2aafddadd9ea0dd/pkg/instance/helper.go#L375

mojtaba-esk:
Closes #531 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new error message for cloning instances that are not in an appropriate state.
	- Added validation to ensure instances can only be cloned if they are in the `StateCommitted` or `StateStopped` states.

- **Documentation**
	- Enhanced comments in the cloning method to clarify cloning behavior, especially concerning state requirements.

- **Tests**
	- Improved test cases for the `Tshark` struct by adding a logger and ensuring proper configuration of the test instance before cloning.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -141,6 +141,10 @@ func (i *Instance) CloneWithSuffix(suffix string) (*Instance, error) {
 // When cloning an instance that is a sidecar, the clone will be not a sidecar
 // When cloning an instance with sidecars, the sidecars will be cloned as well
 func (i *Instance) CloneWithName(name string) (*Instance, error) {
+	if !i.IsInState(StateCommitted, StateStopped) {
+		return nil, ErrCannotCloneInstance.WithParams(i.name, i.state)
+	}
+
 	clonedSidecars, err := i.sidecars.clone(name)
 	if err != nil {
 		return nil, err
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -137,7 +137,7 @@ func (i *Instance) CloneWithSuffix(suffix string) (*Instance, error) {
 }
 
 // CloneWithName creates a clone of the instance with a given name
-// This function can only be called in the state 'Committed'
+// This function can only be called in the state 'Committed' or 'Stopped'
 // When cloning an instance that is a sidecar, the clone will be not a sidecar
 // When cloning an instance with sidecars, the sidecars will be cloned as well
 func (i *Instance) CloneWithName(name string) (*Instance, error) {
--- a/pkg/sidecars/tshark/tshark_test.go
+++ b/pkg/sidecars/tshark/tshark_test.go
@@ -169,7 +169,16 @@ func TestTsharkValidateConfig(t *testing.T) {
 }
 
 func TestTsharkClone(t *testing.T) {
-	testInstance, err := instance.New("testInstance", &system.SystemDependencies{})
+	testInstance, err := instance.New("testInstance",
+		&system.SystemDependencies{
+			Logger: logrus.New(),
+		})
+	require.NoError(t, err)
+
+	err = testInstance.Build().SetImage(context.Background(), "testImage")
+	require.NoError(t, err)
+
+	err = testInstance.Build().Commit(context.Background())
 	require.NoError(t, err)
 
 	tshark := &Tshark{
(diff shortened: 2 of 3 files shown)


pkg/instance/instance.go
```diff
@@ -141,6 +141,10 @@ func (i *Instance) CloneWithSuffix(suffix string) (*Instance, error) {
 // When cloning an instance that is a sidecar, the clone will be not a sidecar
 // When cloning an instance with sidecars, the sidecars will be cloned as well
 func (i *Instance) CloneWithName(name string) (*Instance, error) {
+	if !i.IsInState(StateCommitted, StateStopped) {
+		return nil, ErrCannotCloneInstance.WithParams(i.name, i.state)
+	}
```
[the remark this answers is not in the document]
mojtaba-esk: fixed

mojtaba-esk:
> hello! I found the CI broken, is it expected?
> ...

No it is not expected. something might be wrong.

mojtaba-esk:
When I run the failing test `TestRunSuite/TestBuildFromGit` locally it passes:

```sh
=== RUN   TestRunSuite
time="2024-10-08T15:46:06+03:30" level=info msg="Current log level" file="log/logger.go:41" env_log_level=LOG_LEVEL log_level=info
    suite_setup_test.go:53: Scope: 20241008-154533-860
=== RUN   TestRunSuite/TestBuildFromGit
=== PAUSE TestRunSuite/TestBuildFromGit
=== RUN   TestRunSuite/TestBuildFromGitWithModifications
=== PAUSE TestRunSuite/TestBuildFromGitWithModifications
=== CONT  TestRunSuite/TestBuildFromGit
=== CONT  TestRunSuite/TestBuildFromGitWithModifications
    build_image_test.go:22: Creating new instance
    build_image_test.go:64: Creating new instance
    build_image_test.go:68: Setting git repo
    build_image_test.go:26: Building the image
    build_image_test.go:36: Image built
    build_image_test.go:40: Starting instance
    build_image_test.go:43: Instance started
    build_image_test.go:45: Getting file bytes
=== NAME  TestRunSuite
    suite.go:59: Cleaning up knuu...
    suite.go:63: Knuu is cleaned up
--- PASS: TestRunSuite (54.40s)
    --- PASS: TestRunSuite/TestBuildFromGit (24.49s)
    --- PASS: TestRunSuite/TestBuildFromGitWithModifications (43.83s)
PASS
ok  	github.com/celestiaorg/knuu/e2e/system	98.264s
```

pkg/sidecars/tshark/tshark_test.go
```diff
@@ -169,7 +169,16 @@ func TestTsharkValidateConfig(t *testing.T) {
 }
 
 func TestTsharkClone(t *testing.T) {
```
MSevey, line 171: This update demostrates that our testing is too interwoven. Updates to an `instance.go` file should result in updates to tests in an `instance_test.go` file. Needing to update specific side car tests like tshark to test core functionality is not good. 

Not a blocker for this PR but something we should work towards with this repo. This might require a good amount of mocking architecture to get us to the end goal. 

I think it would be good to resolve this comment into an issue to capture the context/example and then build our an epic from there.

MSevey (approved):

pkg/sidecars/tshark/tshark_test.go
MSevey: Put another way, this repo doesn't have true unit testing, only integration testing.

mojtaba-esk: The main reason that the sidecar "unittest" is updated, is because we updated the instance state check and obviously it has to reject some states which was working before without having those checks. Therefore, we need to update the unittests/e2e tests even some code from external users who try to clone an instance which is not `Committed` nor `Stopped`.

I agree that we do not have proper unit tests. There have been some attempts and we have some unit tests for k8s with acceptable coverage rate. But the instance pkg does not have unit test at the moment.

PR #574, opened 2024-10-15
chore!: change services to be headless

### LINKED ISSUE #514: Change services to be headless
One should use the DNS name of the service and not the IP.

This means that `GetIP` can no longer return the service IP address, as a headless service does not have a cluster IP.

mojtaba-esk:
Closes #514 

## Summary by CodeRabbit

- **New Features**
  - Added methods for retrieving service DNS names and hostnames.
  - Enhanced functionality for service IP and NodePort retrieval.
  - Introduced new error variables to improve error handling related to Kubernetes service management.
  - Added a new `ClientOptions` struct for flexible client configuration, including cluster domain specification.

- **Bug Fixes**
  - Improved error handling with new error variables for service endpoint and IP retrieval restrictions.

- **Refactor**
  - Streamlined network operations by enforcing state checks in IP retrieval methods.
  - Consolidated service creation and patching parameters into a single `ServiceOptions` structure for easier management.
  - Enhanced configuration capabilities by allowing specification of the cluster domain during client initialization.


e2e/system/folder_test.go
```diff
@@ -21,13 +21,12 @@ func (s *Suite) TestFolder() {
 	require.NoError(s.T(), err)
 
 	require.NoError(s.T(), web.Build().Commit(ctx))
+	s.Require().NoError(web.Execution().Start(ctx))
```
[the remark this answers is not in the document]
mojtaba-esk: leave it as it is

e2e/basic/headless_service_test.go
```diff
@@ -0,0 +1,66 @@
...
+	s.Require().NoError(err)
+
+	err = mother.Build().SetImage(ctx, gopingImage)
+	s.Require().NoError(err)
+
+	s.Require().NoError(mother.Network().AddPortTCP(gopingPort))
+	s.Require().NoError(mother.Build().Commit(ctx))
+
+	err = mother.Build().SetEnvironmentVariable("SERVE_ADDR", fmt.Sprintf("0.0.0.0:%d", gopingPort))
+	s.Require().NoError(err)
```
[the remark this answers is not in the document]
smuu: This is not needed as when setting an env variable after the commit, it will be added in the Kubernetes manifest.

pkg/instance/network.go
```diff
@@ -119,34 +119,22 @@ func (n *network) AddPortUDP(port int) error {
 }
 
 // GetIP returns the IP of the instance
-// This function can only be called in the states 'Preparing' and 'Started'
+// This function can only be called in the states 'Started'
 func (n *network) GetIP(ctx context.Context) (string, error) {
```
smuu, line 123: The problem may be that the developer expects this IP to be static. But in fact, the IP changes if the Pod restarts. A restart can happen either by the users' intentions or 'randomly' by Kubernetes.
One idea is to document it properly, but the developer might oversee this. Another option would be to remove `GetIP` and entirely require the switch to `HostName.`

pkg/instance/network.go
```diff
@@ -119,34 +119,22 @@ func (n *network) AddPortUDP(port int) error {
...
 
-	// Update i.kubernetesService for future reference
-	n.kubernetesService = svc
-	return ip, nil
+	return pod.Status.PodIP, nil
+}
+
+func (n *network) HostName() string {
+	return n.instance.K8sClient.ServiceDNS(n.instance.name)
 }
```
smuu, line 138: Very clean 👍

smuu (changes requested):

pkg/instance/network.go
MSevey: We should definitely document the fact that the IP can change.

MSevey (dismissed):
utACK

e2e/system/env_to_json_test.go
```diff
@@ -62,10 +62,7 @@ func (s *Suite) TestEnvToJSON() {
 
 	// Test logic
 	for _, i := range instances {
-		webIP, err := i.Network().GetIP(ctx)
-		s.Require().NoError(err)
-
-		wget, err := executor.Execution().ExecuteCommand(ctx, "wget", "-q", "-O", "-", webIP)
+		wget, err := executor.Execution().ExecuteCommand(ctx, "wget", "-q", "-O", "-", i.Network().HostName())
```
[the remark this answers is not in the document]
mojtaba-esk: no that's fine. they are kept as before to make sure also GetIP works just like before.

pkg/instance/network.go
mojtaba-esk: thanks

pkg/k8s/service.go
```diff
@@ -162,6 +162,11 @@ func (c *Client) WaitForService(ctx context.Context, name string) error {
 	}
 }
 
+func (c *Client) ServiceDNS(name string) string {
+	return fmt.Sprintf("%s.%s.svc.cluster.local", name, c.namespace)
+}
```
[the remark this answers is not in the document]
mojtaba-esk: @smuu how useful do you think making it configurable would be?

mojtaba-esk:
Reminder: It is better to make the creation of headless service optional. i.e. if user wants to create another type of service allow for it. One way could be to have the headless service as default and accept also a custom service.

pkg/instance/network.go
smuu: The function documentation makes sense. But I know the devs will not read that and then get confused that it does not work as expected.
Currently, I don't see any case where they need the IP. Therefore, I vote to remove the `GetIP` function.

pkg/k8s/service.go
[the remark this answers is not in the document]
smuu: If using a default Kubernetes setup, this will not be an issue. There ar setups where the cluster domain is different.
One option would be to add this as an optional parameter of the knuu instance with `cluster.local` as the default value.

smuu (changes requested):

pkg/instance/network.go
MSevey: There is also GetServiceIP. 

Removing one or both of those methods would create a lot of bloat to this PR. 

If we want to remove that I recommend we do that in a separate PR.

MSevey (dismissed):

pkg/instance/network.go
smuu: `GetServiceIP` can't return an IP if it's `headless`.

mojtaba-esk: Maybe we can change the `GetIP` name to something else like `GetEphemeralIP` and for `GetServiceIP`, i think there is a specific error when it is headless, if not I will add it.

mojtaba-esk: I checked it, there is a particular error when user requests GetServiceIP on a headless service

pkg/k8s/service.go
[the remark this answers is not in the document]
mojtaba-esk: that's a good idea. thanks guys, will add it.

pkg/k8s/service.go
```diff
@@ -206,16 +243,63 @@
...
+		return false, ErrGettingPodsForService.WithParams(service.Name).Wrap(err)
+	}
+	// Check if at least one pod is ready
+	for _, pod := range pods.Items {
+		if pod.Status.Phase == v1.PodRunning {
+			return true, nil
+		}
+	}
+	return false, nil
+}
```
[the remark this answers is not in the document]
mojtaba-esk: thanks. referenced in a new issue

pkg/k8s/service_test.go
```diff
@@ -464,7 +482,7 @@ func (s *TestSuite) TestWaitForService() {
 				defer listener.Close()
 			}
 
-			ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
+			ctx, cancel := context.WithTimeout(context.Background(), 200*time.Microsecond)
```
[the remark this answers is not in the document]
mojtaba-esk: nice catch!

PR #580, opened 2024-11-14
removed names.NewRandomK8() from the project completely

### LINKED ISSUE #573: remove `names.NewRandomK8()` completely from the project
The `names.NewRandomK8()` function which introduces randomness in the k8s resource names, are still being used in some parts of the project such as Traefik reverse proxy pkg. 

In order to have long running tests, we need to remove them completely.

notlelouch:
closes #573  

## Release Notes

- **New Features**
  - Enhanced job name generation for Kaniko builds using the current Unix timestamp for better job identification in Kubernetes.
  - Simplified temporary naming for instances and Kubernetes resources using timestamps, improving uniqueness and readability.

- **Bug Fixes**
  - Improved error handling across various methods, providing clearer error messages for different failure scenarios.

- **Chores**
  - Removed the dependency on the external naming package, streamlining the naming process for service accounts and roles.


pkg/builder/kaniko/kaniko.go
```diff
@@ -133,10 +133,7 @@ func (k *Kaniko) containerLogs(ctx context.Context, pod *v1.Pod) (string, error)
 }
 
 func (k *Kaniko) prepareJob(ctx context.Context, b *builder.BuilderOptions) (*batchv1.Job, error) {
-	jobName, err := names.NewRandomK8(kanikoJobNamePrefix)
-	if err != nil {
-		return nil, ErrGeneratingUUID.Wrap(err)
-	}
+	jobName := fmt.Sprintf("%s-%d", kanikoJobNamePrefix, time.Now().UnixNano())
```
MSevey, line 136: This is still random from the perspective that the user is not defining the name.

The intention behind the issue is that all names should be explicitly set by the user so that they have knowledge of them in order to reference and update things.

MSevey (changes requested):
Thanks for submitting a PR!

The intention of the issue was to replace the `NewRandomK8` function with the user passing in the name as an argument. 

So for every function that had the the `NewRandomK8` function being called, you have to add the `jobName` as an input and propagate that up through the call stack.

pkg/builder/kaniko/kaniko.go
mojtaba-esk: it does not matter for kaniko as it creates a temp job to build and does not affect the long running tests.

Issue #529, opened 2024-07-25
Have files as a separate object


smuu:
Currently, files are added to instances with instance.AddFile(size).

I propose to add a File object, that can be added to an Instance with instance.AddFile(file).

Related: https://github.com/celestiaorg/knuu/issues/390

Issue #390, opened 2024-05-29
Have volumes as a separate object


smuu:
Currently, volumes are added to instances with `instance.AddVolume(size)`.

I propose to add a `Volume` object, that can be added to an Instance with `instance.AddVolume(volume)`.
This enables adding the same volume to multiple containers in one sidecar.

smuu:
Related: https://github.com/celestiaorg/knuu/issues/167

smuu:
Related: https://github.com/celestiaorg/knuu/issues/317

PR #589, opened 2024-12-04
feat(storage): retain file permission and owner when adding via confi…

smuu:
…gMap

## Release Notes

- **New Features**
  - Enhanced file handling in Kubernetes pods with new attributes for ownership and permissions.
  - Dynamic generation of build directories for improved flexibility.

- **Bug Fixes**
  - Improved error handling for file operations, providing specific error messages for better diagnostics.

- **Refactor**
  - Streamlined file deployment and destruction processes to prevent unnecessary operations.
  - Updated method signatures to reflect changes in parameters related to file ownership and permissions.

- **Chores**
  - Updated OpenTelemetry agent configuration file permissions for enhanced security.


pkg/instance/build.go
```diff
@@ -226,7 +225,11 @@ func getImageRegistry(imageName string) (string, error) {
 
 // getBuildDir returns the build directory for the instance
 func (b *build) getBuildDir() string {
-	return filepath.Join(buildDirBase, b.instance.name)
+	tmpDir, err := os.MkdirTemp("", "knuu-build-*")
```
MSevey, line 228: why the `*`?

pkg/k8s/pod.go
```diff
@@ -442,11 +467,25 @@ func buildInitContainerVolumes(name string, volumes []*Volume, files []*File) []
 		return []v1.VolumeMount{} // return empty slice if no volumes are specified
 	}
 
-	containerVolumes := []v1.VolumeMount{
-		{
+	var containerVolumes []v1.VolumeMount
+	if len(volumes) != 0 {
```
MSevey, line 471: It feels like we should have some context captured in comments as to the split out of these if conditions. 

I don't understand why we are checking `len(volumes)` but not doing anything with it in either if condition.

pkg/k8s/pod.go
```diff
@@ -414,25 +425,39 @@ func buildContainerVolumes(name string, volumes []*Volume, files []*File) []v1.V
...
-	var containerFiles []v1.VolumeMount
-
-	for n, file := range files {
-		shouldAddFile := true
-		for _, volume := range volumes {
-			if strings.HasPrefix(file.Dest, volume.Path) {
-				shouldAddFile = false
-				break
-			}
+	if len(volumes) == 0 && len(files) != 0 {
```
MSevey, line 428: what's the main difference between `buildContainerVolumes` and `buildInitContainerVolumes`? 

Feels like we should be able to refactor this to something like:

```go
func buildInitContainerVolumes() {
  // some code
 return buildContainerVolumes()
}
```

MSevey: maybe a good follow up. Looks like there are a bunch of functions in this file that could be structured in this way.

pkg/instance/build.go
```diff
@@ -226,7 +225,11 @@ func getImageRegistry(imageName string) (string, error) {
...
+	if err != nil {
+		return ""
+	}
+	return filepath.Join(tmpDir, b.instance.name)
```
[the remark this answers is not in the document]
mojtaba-esk: This comment is valid, let's handle that error.

pkg/instance/errors.go
```diff
@@ -102,7 +102,9 @@ var (
 	ErrCreatingDirectory                         = errors.New("CreatingDirectory", "error creating directory")
 	ErrFailedToCreateDestFile                    = errors.New("FailedToCreateDestFile", "failed to create destination file '%s'")
 	ErrFailedToOpenSrcFile                       = errors.New("FailedToOpenSrcFile", "failed to open source file '%s'")
+	ErrFailedToGetSrcFileInfo                    = errors.New("FailedToGetSrcFileInfo", "failed to get source file info for")
```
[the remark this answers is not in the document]
mojtaba-esk: Also this one `%s` is required for the `WithParam` to work.

pkg/instance/storage.go
```diff
@@ -249,46 +248,55 @@ func (s *storage) copyFileToBuildDir(src, dest string) (string, error) {
...
+		return "", ErrFailedToCreateDestFile.WithParams(dstPath).Wrap(err)
+	}
+	defer dst.Close()
+
 	if _, err := io.Copy(dst, srcFile); err != nil {
 		return "", ErrFailedToCopyFile.WithParams(src, dstPath).Wrap(err)
 	}
 
+	// Ensure the destination file has the same permissions as the source file
+	if err := os.Chmod(dstPath, srcInfo.Mode().Perm()); err != nil {
```
mojtaba-esk, line 273: [nit] When we create a file with `srcInfo.Mode().Perm()` do we really need to apply again the `chmod`? Is this one tested that the `OpenFile` sometimes fails and we need to make sure on it?

pkg/instance/storage.go
```diff
@@ -249,46 +248,55 @@ func (s *storage) copyFileToBuildDir(src, dest string) (string, error) {
...
 	parts := strings.Split(chown, ":")
 	if len(parts) != 2 {
-		return ErrInvalidFormat
+		return ErrInvalidFormat.WithParams(chown)
 	}
-
-	group, err := strconv.ParseInt(parts[1], 10, 64)
-	if err != nil {
-		return ErrFailedToConvertToInt64.Wrap(err)
+	for _, part := range parts {
```
mojtaba-esk, line 293: [nit] we could also use a regex here to validate the chown input.

pkg/k8s/pod.go
```diff
@@ -414,25 +425,39 @@ func buildContainerVolumes(name string, volumes []*Volume, files []*File) []v1.V
...
+	// 	}
+	// 	if shouldAddFile {
+	// 		containerFiles = append(containerFiles, v1.VolumeMount{
+	// 			Name:      name + podFilesConfigmapNameSuffix,
+	// 			MountPath: file.Dest,
+	// 			SubPath:   fmt.Sprintf("%d", n),
+	// 		})
+	// 	}
+	// }
+
```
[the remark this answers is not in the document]
mojtaba-esk: we should not have commented code here. They should be either uncommented or removed.

pkg/k8s/pod.go
```diff
@@ -483,26 +522,33 @@ func (c *Client) buildInitContainerCommand(volumes []*Volume, files []*File) []s
 			cmds = append(cmds, parentDirCmd)
 			dirsProcessed[folder] = true
 		}
-		copyFileToKnuu := fmt.Sprintf("cp %s %s && ", file.Dest, filepath.Join(knuuPath, file.Dest))
-		cmds = append(cmds, copyFileToKnuu)
+		chown := file.Chown
+		permission := file.Permission
+		addFileToKnuu := fmt.Sprintf("cp %s %s && ", file.Dest, filepath.Join(knuuPath, file.Dest))
+		if chown != "" {
```
mojtaba-esk, line 528: why not use  the `file.Chown` itself, do we change it somewhere? same for permission

pkg/instance/build.go
mojtaba-esk: you mean why `b *build` ? 
if so, we need to pass it by reference because it is more performant than not using it which copies the entire object. Also we need it as there is a `*sync.Map` in it, and this way it works properly.

mojtaba-esk (changes requested):
Great work!!
left some nit comments and suggestions.

pkg/instance/storage.go
```diff
@@ -130,6 +130,9 @@ func (s *storage) AddFileBytes(bytes []byte, dest string, chown string) error {
 	if _, err := tmpfile.Write(bytes); err != nil {
 		return err
 	}
+	if err := tmpfile.Chmod(0644); err != nil {
```
mojtaba-esk, line 133: [nit] would be nicer to extract the `0644` constant

pkg/k8s/pod.go
```diff
@@ -594,7 +595,7 @@ func (c *Client) prepareInitContainers(config ContainerConfig, init bool) []v1.C
 	return []v1.Container{
 		{
 			Name:  config.Name + initContainerNameSuffix,
-			Image: config.Image,
+			Image: initContainerImage,
```
mojtaba-esk, line 598: [Q] why does it have to be a fixed image?

pkg/k8s/pod.go
```diff
@@ -619,9 +620,11 @@ func (c *Client) preparePodSpec(spec PodConfig, init bool) v1.PodSpec {
 
 	// Prepare sidecar containers and append to the pod spec
 	for _, sidecarConfig := range spec.SidecarConfigs {
+		sidecarInitContainer := c.prepareInitContainers(sidecarConfig, true)
```
mojtaba-esk, line 623: [Q] how many init containers a pod can have? I thought it could be only one.

mojtaba-esk (approved):
Nice work. Overall LGTM, I had some questions which I left in the code.

pkg/k8s/pod.go
```diff
@@ -71,8 +68,10 @@ type Volume struct {
 }
 
 type File struct {
-	Source string
-	Dest   string
+	Source     string
+	Dest       string
+	Chown      string
+	Permission string
```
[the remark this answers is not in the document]
smuu: This is validated when adding a file.

pkg/k8s/pod.go
```diff
@@ -101,10 +100,12 @@
...
 
-func (c *Client) NewFile(source, dest string) *File {
+func (c *Client) NewFile(source, dest, chown, permission string) *File {
 	return &File{
-		Source: source,
-		Dest:   dest,
+		Source:     source,
+		Dest:       dest,
+		Chown:      chown,
+		Permission: permission,
```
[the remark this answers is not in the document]
smuu: This is only used internally, no documentation needed

pkg/instance/build.go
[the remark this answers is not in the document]
smuu: Fixed with https://github.com/celestiaorg/knuu/pull/589/commits/10bedaf8ee2dc3e0d6cbd85f9130f4ba19f39b49

pkg/instance/storage.go
```diff
@@ -249,46 +248,55 @@ func (s *storage) copyFileToBuildDir(src, dest string) (string, error) {
...
-
-	group, err := strconv.ParseInt(parts[1], 10, 64)
-	if err != nil {
-		return ErrFailedToConvertToInt64.Wrap(err)
+	for _, part := range parts {
+		if _, err := strconv.ParseInt(part, 10, 64); err != nil {
+			return ErrFailedToConvertToInt64.WithParams(part).Wrap(err)
+		}
 	}
+	file := s.instance.K8sClient.NewFile(srcPath, dest, chown, permission)
```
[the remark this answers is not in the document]
smuu: Before calling `addFileToInstance` the chown string is validated:

```
// validateFileArgs validates the file arguments
func (s *storage) validateFileArgs(src, dest, chown string) error {
	if src == "" {
		return ErrSrcMustBeSet
	}
	if dest == "" {
		return ErrDestMustBeSet
	}
	if chown == "" {
		return ErrChownMustBeSet
	}

	// validate chown format
	if !strings.Contains(chown, ":") || len(strings.Split(chown, ":")) != 2 {
		return ErrChownMustBeInFormatUserGroup
	}
	return nil
}
```

pkg/k8s/pod.go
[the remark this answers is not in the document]
smuu: Fixed in: https://github.com/celestiaorg/knuu/pull/589/commits/57302c1a40338fdbdd81dc81cf4154cced329264

pkg/instance/build.go
smuu:
```
func os.MkdirTemp(dir string, pattern string) (string, error)
MkdirTemp creates a new temporary directory in the directory dir and returns the pathname of the new directory. The new directory's name is generated by adding a random string to the end of pattern. If pattern includes a "*", the random string replaces the last "*" instead. The directory is created with mode 0o700 (before umask). If dir is the empty string, MkdirTemp uses the default directory for temporary files, as returned by TempDir. Multiple programs or goroutines calling MkdirTemp simultaneously will not choose the same directory. It is the caller's responsibility to remove the directory when it is no longer needed.
```

pkg/k8s/pod.go
smuu: What do you mean by not doing anything?

But I added some comments to explain why we need it like that.

See: https://github.com/celestiaorg/knuu/pull/589/commits/08c587ebe34f72819484a0c80df66a6165baaa90

pkg/k8s/pod.go
smuu: Even if the functions look similar, they do not do the same. The init container is used to prepare files and volumes for the main container.

I agree that we need to improve file and volume handling.

Ref.:
- https://github.com/celestiaorg/knuu/issues/390
- https://github.com/celestiaorg/knuu/issues/529

pkg/instance/storage.go
smuu: While developing it did not properly apply the correct permission. Because of this, I added this to ensure the permission is correct.

pkg/k8s/pod.go
smuu: What do you mean?
When mounting a file through a configMap, it does not preserve file owner and permissions from the original file. A ConfigMap has no knowledge about ownership or permissions. This is why we need to mount the ConfigMap in the init container, change the owner and permission, and then move it to a volume that is mounted to the main container.

pkg/instance/storage.go
smuu: thank you!
https://github.com/celestiaorg/knuu/pull/589/commits/34a7443782f68347cc7e267ec868a0ee94e8d5f5

pkg/k8s/pod.go
smuu: When it is the same image as the main container, the init container can fail.
The init container needs to create directories, move files, and change the owner and permissions of files. As not each application container has those tools, we need to use an image in the init container that has these tools.

pkg/k8s/pod.go
smuu: There is no limit to the init container.
For example, if you have a main container and a sidecar (eg. otel-agent), then we now have 2 init containers.

pkg/k8s/pod.go
mojtaba-esk: I meant why we assign it to a new variable. we could have it like so:

```go
if file.Chown != "" {
   ...
```

pkg/instance/storage.go
mojtaba-esk: so then we can remove these lines:

```go
for _, part := range parts {
   if _, err := strconv.ParseInt(part, 10, 64); err != nil {
		return ErrFailedToConvertToInt64.WithParams(part).Wrap(err)
	}
}
```

mojtaba-esk (approved):
well done, left few nits

pkg/instance/storage.go
smuu: fixed

mojtaba-esk (approved):

pkg/instance/storage.go
```diff
@@ -291,30 +331,23 @@
...
~	parts := strings.Split(chown, ":")
~	if len(parts) != 2 {
 		return ErrInvalidFormat
 	}
 
-	group, err := strconv.ParseInt(parts[1], 10, 64)
-	if err != nil {
-		return ErrFailedToConvertToInt64.Wrap(err)
-	}
+	file := s.instance.K8sClient.NewFile(srcPath, dest, chown, permission)
```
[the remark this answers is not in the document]
smuu: fixed

PR #598, opened 2024-12-19
A simple long running test

mojtaba-esk:
This PR proposes a simple long running test. Some of the underlying code had to change to allow for that.

## Summary by CodeRabbit

- **New Features**
	- Introduced new end-to-end tests for instance management and command execution.
	- Enhanced logging for file operations in the storage module.
	- Added a method to create ClusterRoleBinding resources in the test suite.
	- Added a `skipCleanup` option to control cleanup behavior in the Knuu struct.
	- Added a new test function for validating image upgrades in long-running instances.
	- Introduced utility functions for creating and starting instances.
	- Added new error variables for handling existing Kubernetes resources.

- **Bug Fixes**
	- Improved error handling for existing Kubernetes resources in various methods.
	- Added specific error messages for existing service accounts, roles, and role bindings.

- **Refactor**
	- Updated various methods to use a more modern and fluent API for Kubernetes probes and configurations.
	- Transitioned to a declarative approach for managing Kubernetes resources.
	- Enhanced error handling logic across multiple components for better clarity and robustness.

- **Chores**
	- Cleaned up indirect dependencies in the `go.mod` file and updated several module versions.


smuu:
Is this PR based on another PR? I can see many changes here and not only adding a test.
What files should I review here?

mojtaba-esk:
> Is this PR based on another PR? I can see many changes here and not only adding a test. What files should I review here?

No I don't think so, if I recall well, to run the test, some parts of the knuu code had to change.

tty47 (dismissed):

pkg/instance/storage.go
```diff
@@ -342,6 +342,9 @@ func (s *storage) addFileToInstance(srcPath, dest, chown string) error {
 
 	file := s.instance.K8sClient.NewFile(srcPath, dest, chown, permission)
 
+	// TODO: remove this
+	fmt.Printf("\nfile: %#v\n", file)
+
```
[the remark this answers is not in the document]
smuu: I think it would be good to have this as a debug message instead of a print

e2e/longrunning/simple_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/celestiaorg/knuu/pkg/instance"
+	"github.com/celestiaorg/knuu/pkg/knuu"
+)
+
+const (
+	alpineImage = "alpine:3.20.3"
+	testTimeout = time.Minute * 5
+)
+
+func TestSimple(t *testing.T) {
```
smuu, line 19: Can we add a test where we test to change something (for example, the image) of an already running instance

e2e/longrunning/simple_test.go
smuu, line 1: Not blocking for this PR, but I'm curious.

How would it look if I wanted to delete an instance from a long-running test?

smuu (changes requested):

e2e/longrunning/simple_test.go
mojtaba-esk: Genius question, I will think about it.

e2e/longrunning/simple_test.go
mojtaba-esk: yes will add it

e2e/longrunning/simple_test.go
mojtaba-esk: I believe user can just use the `Destroy()` method and it should work.

e2e/longrunning/simple_test.go
mojtaba-esk: yes it is added now. There was an issue with pvc applying and that one is also resolved in the recent commit

tty47 (dismissed):

MSevey:
Do we still plan on supporting long running tests with knuu now that we are working on Talis? 
If not, is this PR still applicable?

mojtaba-esk:
> Do we still plan on supporting long running tests with knuu now that we are working on Talis? If not, is this PR still applicable?

I think we are not going to sunset knuu anytime soon. so this might be used til then. @smuu knows more accurately.

e2e/longrunning/simple_test.go
MSevey: I would make this a blocking change request. 

The key features of a long running test would be to update a node and remove nodes.  So without that, this PR is incomplete. 

There also just needs to be one test, I don't think the `TestSimple` is providing value on its own. 

It should just be `TestLongRunning` with confirms the core features of a long running test. 
- Launch 2 nodes
- Perform readiness checks
- update one node
- confirm update
- delete the other node
- verify deletion
- confirm other node is still operating as intended

I don't think it needs the 5s sleep as the length of time is not meaningful in the testing context.

MSevey (changes requested):

PR #600, opened 2025-01-07
feat: k8s auth token instead of kube conffig

### LINKED ISSUE #319: feat: Authentication with token
Instead of requiring a kubeconfig the DevEx would improve when only needing a token to authenticate against the k8s cluster.

Ref.: https://kubernetes.io/docs/reference/access-authn-authz/authentication/

mojtaba-esk:
Closes #319 

## Summary by CodeRabbit

- **New Features**
	- Enhanced Kubernetes client configuration with flexible options.
	- Added logging capabilities to test suites.
	- Introduced authentication token generation script for Kubernetes access.

- **Bug Fixes**
	- Improved error handling for Kubernetes client initialization.
	- Added checks for cluster configuration before executing commands.

- **Chores**
	- Refactored Kubernetes client setup across multiple test suites.
	- Added new utility functions for managing Kubernetes client options.

### DIFF
--- a/e2e/suite.go
+++ b/e2e/suite.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"os"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -11,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/celestiaorg/knuu/pkg/instance"
+	"github.com/celestiaorg/knuu/pkg/k8s"
 	"github.com/celestiaorg/knuu/pkg/knuu"
 )
 
@@ -92,4 +94,19 @@ func (s *Suite) RetryOperation(operation func() error, maxRetries int) error {
 		time.Sleep(time.Second * time.Duration(i+1))
 	}
 	return fmt.Errorf("operation failed after %d retries: %w", maxRetries, err)
+}
+
+func (s *Suite) K8sDefaultOptions() []k8s.Option {
+	if os.Getenv("K8S_AUTH_TOKEN") == "" {
+		s.T().Log("K8S_AUTH_TOKEN is not set, using default cluster config from ~/.kube/config")
+		return nil
+	}
+
+	return []k8s.Option{
+		k8s.WithAuthToken(
+			os.Getenv("K8S_HOST"),
+			os.Getenv("K8S_CA_CERT"),
+			os.Getenv("K8S_AUTH_TOKEN"),
+		),
+	}
 }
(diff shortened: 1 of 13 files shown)


scripts/auth_token.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+# This script generates a bearer token for the knuu service account.
+# It is used to authenticate the knuu service account to the Kubernetes API server.
+# The token is used to test things manually. 
+# It requries ~/.kube/config to be set up.
+
+# Variables
+SERVICE_ACCOUNT_NAME="knuu-service-account"
+NAMESPACE="default"
+ROLE_BINDING_NAME="knuu-rolebinding"
```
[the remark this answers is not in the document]
mojtaba-esk: ignore that bash script as it is only used in manual testing the idea and we might remove it in near future

e2e/suite.go
```diff
@@ -93,3 +99,18 @@ func (s *Suite) RetryOperation(operation func() error, maxRetries int) error {
 	}
 	return fmt.Errorf("operation failed after %d retries: %w", maxRetries, err)
 }
+
+func (s *Suite) K8sDefaultOptions() []k8s.Option {
+	if os.Getenv(envK8sAuthToken) == "" || os.Getenv(envK8sCACert) == "" || os.Getenv(envK8sHost) == "" {
+		s.T().Logf("%s, %s and/or %s are not set, using default cluster config from ~/.kube/config", envK8sAuthToken, envK8sCACert, envK8sHost)
+		return nil
+	}
```
MSevey, line 107: If i'm understanding the requirements here correctly, I think the logic should be if `ALL` are blank, then use the default cluster config. But if `SOME` are blank, then it should be an error that the developer didn't provider everything. 

If it in fact is OK to provide some, but not all, then the `if` condition definitely needs to be `AND` checks.

pkg/k8s/clinet_options.go
MSevey, line 1: I think the file name is misspelled, should be `client_options.go`

pkg/k8s/clinet_options.go
```diff
@@ -0,0 +1,43 @@
...
+
+type ClientOptions struct {
+	clusterDomain string
+	clusterHost   string
+	authToken     string
+	cert          string
+	clusterConfig *rest.Config
+}
+
+type Option func(*ClientOptions)
```
MSevey, line 13: Didn't we have this discussion with the KNUU options? It's less code to maintain to just have a `New` and `NewWithOptions` function that accepts the options params as opposed to having a unique function call for each option. Then all the input validation can be contained in one place too instead of across all the individual option functions.

scripts/auth_token.sh
MSevey, line 1: This should probably be added to the README right?

MSevey (changes requested):
Some clean up and documentation edits.

